PR #22, opened 2023-05-27
Added requirements.txt.

cvarrichio:
I've re-added a requirements.txt file.  This works for me - I've never gotten the pipenv to work, and I think we should offer this as an alternative.


FraserLee:
Since the situation has evolved so much since this was opened, I'm closing the PR. If this is something we want to revisit lets open another issue.

PR #27, opened 2023-06-09
Add a default question when opening the page for the first time

FraserLee:
@ccstan99 @henri123lemoine @Thomas-Lemoine and anyone else with an option, I want some feedback on this. 

It seems from looking at interactions that people freeze when asked for a random question, or ask something like "what's this?" - not very helpful with the current chatbot model. This could launch them down the right path more often.

### DIFF
--- a/web/src/pages/index.tsx
+++ b/web/src/pages/index.tsx
@@ -460,10 +460,10 @@ const Home: NextPage = () => {
             <main>
                 <Header page="index" />
 
-                <p><b>
+                <p>
                   Since this is still an early test, all questions and answers are stored.<br/>
                   Zero other information is collected.
-                </b></p>
+                </p>
 
                 <ul>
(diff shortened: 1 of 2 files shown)


FraserLee:
The question I have right now is "What is an AI arms race?", but we might want to randomly cycle between a few or something

FraserLee:
it would also be neat if we were able to answer "What is stampy chat?". Though maybe not helpful as inspiration for continuing the convo on their own.

ccstan99:
The ones @plexish and others suggested during the meeting were meant to gauge the users background & interest, along the lines of:
1. What is AI safety and alignment?
2. Why is this important?
3. What are some of the different research approaches?
4. How can I help with AI safety and alignment?

plexish:
Those fit well on the FAQ system, not sure if they're ideal on the Chat though, maybe an example question is better?

PR #38, opened 2023-06-21
ChunkedARD + Embeddings from huggingface's ARD, simplified

Thomas-Lemoine:
Now uses the huggingface dataset https://huggingface.co/datasets/StampyAI/alignment-research-dataset/ instead of local .jsonl files to construct the ChunkedARD. 

Code is simplified in many places, and now assumes some formatting from the cleaned up alignment-research-dataset. 

Prelimiary feature necessary for, later on, modifying pinecone based on whether an id is already in the pinecone db or not.

It currently does not use the planned database of all metadata values for chunks, and thus many fields are ignored, like the tags in LW, the book_title in agentmodels, etc.

But overall it's much less messy than it was.

### DIFF
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -65,8 +65,8 @@ class ChunkedARD:
             'title': str,
             'url': str,
             'date_published': str,
-            'authors': list, # It appears to be a list, but actually it's a list-looking string. Like "['apple', 'orange', 'tomato']" is a string.
-#            'summary': str  # see previous comment
+            'authors': list, # unsure as of yet if this is a string or a list. TODO it is subject to change
+#           'summary': list  # see previous comment
         }
         required_metadata_keys = ['id', 'source', 'title', 'text']
 
@@ -115,6 +115,7 @@ class ChunkedARD:
             url: str = entry['url']
             date_published: str = entry['date_published']
             authors: list = entry['authors']
+            # summary is ignored for now, since most sources lack one. TODO: add summary. see self.metadata code as well.
             
             # Get signature
             if authors:
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -25,16 +25,6 @@ from .settings import PATH_TO_DATASET_DICT_PKL, EMBEDDING_MODEL, LEN_EMBEDDINGS
 from .text_splitter import TokenSplitter, split_into_sentences
 
 
-
-error_count_dict = {
-    "Entry has no id.": 0,
-    "Entry has no source.": 0,
-    "Entry has no title.": 0,
-    "Entry has no text.": 0,
-    "Entry has non-string required key.": 0
-}
-
-
 class MissingDataException(Exception):
     pass
 
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -11,7 +11,6 @@ from tqdm.auto import tqdm
 from dateutil.parser import parse, ParserError
 import openai
 from datasets import load_dataset
-from langchain.document_loaders import HuggingFaceDatasetLoader
 
 
 try:
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -12,6 +12,8 @@ from dateutil.parser import parse, ParserError
 import openai
 from datasets import load_dataset
 
+import logging
+logger = logging.getLogger(__name__)
 
 try:
     import config
@@ -103,8 +105,8 @@ class ChunkedARD:
             try:
                 self._validate_required_metadata(entry)
             except MissingDataException as mde:
-                #logging.error(str(mde))  # Log the error message
-                print(str(mde))
+                logging.error(str(mde))  # Log the error message
+                continue
             except Exception as e:
                 raise e
 
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -74,7 +74,7 @@ class ChunkedARD:
 
         # Check that the 8 primary metadata keys all have the correct type and the key exists
         for key, key_type in metadata_types.items():
-            if not entry[key]:
+            if not key in entry.keys():
                 raise MissingDataException(f"Entry {entry['id']} has no {key}.")
             
             if not isinstance(entry[key], key_type):
@@ -109,11 +109,6 @@ class ChunkedARD:
                 continue
             except Exception as e:
                 raise e
-
-
-            #if the text is too short, ignore this text
-            if len(entry['text']) < 500:
-                continue
 
 
             """Checks are done, so we construct the metadata."""
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -74,7 +74,7 @@ class ChunkedARD:
 
         # Check that the 8 primary metadata keys all have the correct type and the key exists
         for key, key_type in metadata_types.items():
-            if not key in entry.keys():
+            if not key in entry:
                 raise MissingDataException(f"Entry {entry['id']} has no {key}.")
             
             if not isinstance(entry[key], key_type):
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -91,7 +91,7 @@ class ChunkedARD:
     def get_alignment_texts(self):
         text_splitter = TokenSplitter(self.min_tokens_per_block, self.max_tokens_per_block)
 
-        # Load the dataset. streaming allows you to load one entry at a time, 
+        # Load the dataset as an iterable.
         iterable_data = self.iterable_data(self)
 
         for entry in iterable_data:            
@@ -126,7 +126,7 @@ class ChunkedARD:
             url: str = entry['url']
             date_published: str = entry['date_published']
             authors: list = entry['authors']
-            # summary is ignored for now, since most sources lack one. TODO: add summary. see self.metadata code as well.
+            summary: list = entry['summary']
             
             # Get signature
             if authors:
@@ -147,7 +147,8 @@ class ChunkedARD:
                     'text': chunk,
                     'url': url,
                     'date_published': date_published,
-                    'authors': authors
+                    'authors': authors,
+                    'summary': summary
                 })
             
             # Update counts
@@ -159,12 +160,12 @@ class ChunkedARD:
             self.total_counts['chunks'] += len(chunks)
 
     def show_stats(self):
-        print(f'Number of entries by source: {self.entries_per_source_count}')
-        print(f'Total entries count: {self.total_counts["entries"]}')
-        print(f'Total character count: {self.total_counts["chars"]}')
-        print(f'Total word count: {self.total_counts["words"]}')
-        print(f'Total sentence count: {self.total_counts["sentences"]}')
-        print(f'Total chunk count: {self.total_counts["chunks"]}')
+        logger.info(f'Number of entries by source: {self.entries_per_source_count}')
+        logger.info(f'Total entries count: {self.total_counts["entries"]}')
+        logger.info(f'Total character count: {self.total_counts["chars"]}')
+        logger.info(f'Total word count: {self.total_counts["words"]}')
+        logger.info(f'Total sentence count: {self.total_counts["sentences"]}')
+        logger.info(f'Total chunk count: {self.total_counts["chunks"]}')
 
     def get_embeddings(self):
         def get_embeddings_at_index(texts: str, batch_idx: int, batch_size: int = 200): # int, np.ndarray
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -97,7 +97,9 @@ class ChunkedARD:
 
         # Load the dataset. streaming allows you to load one entry at a time, 
         # so entries can be processed before the entire dataset has been saved.
-        iterable_data = load_dataset('StampyAI/alignment-research-dataset', 'aisafety.info', split='train', streaming=True)
+        # iterable_data = load_dataset('StampyAI/alignment-research-dataset', 'aisafety.info', split='train', streaming=True)
+
+        iterable_data = load_dataset('StampyAI/alignment-research-dataset', 'all', split='train', streaming=True)
         
         for entry in tqdm(iterable_data):            
             """Checks"""
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -92,7 +92,7 @@ class ChunkedARD:
         text_splitter = TokenSplitter(self.min_tokens_per_block, self.max_tokens_per_block)
 
         # Load the dataset as an iterable.
-        iterable_data = self.iterable_data(self)
+        iterable_data = self.iterable_data()
 
         for entry in iterable_data:            
             """Checks"""
--- a/src/dataset/create_dataset.py
+++ b/src/dataset/create_dataset.py
@@ -1,4 +1,3 @@
-import jsonlines
 import numpy as np
 from typing import List, Dict, Tuple, DefaultDict, Any
 from collections import defaultdict
@@ -11,6 +10,9 @@ from pathlib import Path
 from tqdm.auto import tqdm
 from dateutil.parser import parse, ParserError
 import openai
+from datasets import load_dataset
+from langchain.document_loaders import HuggingFaceDatasetLoader
+
 
 try:
     import config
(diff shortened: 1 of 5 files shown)


src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
-            raise MissingDataException(f"Entry has no text.")
+    def contains_required_metadata(self, entry: Dict[str, Any]):
+        metadata_types = {
+            'id': str,
+            'source': str,
+            'title': str,
+            'url': str,
+            'date_published': str,
+            'authors': list, # unsure as of yet if this is a string or a list. TODO it is subject to change
+#           'summary': list  # see previous comment
```
mruwnik, line 69: assume that it's a list of strings and that it will always be present (if not there's a bug in ARD)

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+        }
+        required_metadata_keys = ['id', 'source', 'title', 'text']
+
+        # Check that the 8 primary metadata keys all have the correct type
+        for key, key_type in metadata_types.items():
+            if type(entry[key]) != key_type:
```
mruwnik, line 75: `if not isinstance(entry[key], key_type)`

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
-            if type(article['tags']) == list: tags = ', '.join([val['term'] for val in article['tags']])
-            elif type(article['tags']) == str: tags = article['tags']
-            else: tags = None
+                self.entries_per_source_count[source] = 0
         
-        # Get text
-        if 'text' in article and article['text']: text = article['text']
-        else:
-            raise MissingDataException(f"Entry has no text.")
+    def contains_required_metadata(self, entry: Dict[str, Any]):
```
mruwnik, line 61: [NIT] maybe `validate_required_metadata`? I have the feeling that function that are named like `contains_`, `is_` etc. tend to return a boolean, while functions which will raise an exception if something is wrong tend to be called `validate_`

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+        }
+        required_metadata_keys = ['id', 'source', 'title', 'text']
+
+        # Check that the 8 primary metadata keys all have the correct type
+        for key, key_type in metadata_types.items():
+            if type(entry[key]) != key_type:
+                raise MissingDataException(f"Entry {entry['id']} has key {key} of type {type(entry[key])} when it should be {key_type}.")
+
+        # Check that the 4 required metadata keys are non-empty
+        for key in required_metadata_keys:
```
mruwnik, line 79: just do `for key in metadata_types` - all the keys should be present

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
-                except MissingDataException as e:
-                    if str(e) not in error_count_dict:
-                        error_count_dict[str(e)] = 0
-                    error_count_dict[str(e)] += 1
+
+        # Load the dataset. streaming allows you to load one entry at a time, 
+        # so entries can be processed before the entire dataset has been saved.
+        # iterable_data = load_dataset('StampyAI/alignment-research-dataset', 'aisafety.info', split='train', streaming=True)
+
+        iterable_data = load_dataset('StampyAI/alignment-research-dataset', 'all', split='train', streaming=True)
```
mruwnik, line 91: how about extracting this to a method (or even property), which would be something like:
```
def iterable_data(self):
    if not self.custom_sources:
        return tqdm(load_dataset('StampyAI/alignment-research-dataset', 'all', split='train', streaming=True))
    
    data = (entry for source in self.custom_sources for entry in load_dataset('StampyAI/alignment-research-dataset', source, split='train', streaming=True))
    return tqdm(data)
```

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+        
+        for entry in tqdm(iterable_data):            
+            """Checks"""
+
+            # if we specified custom sources, only include entries from those sources
+            if (self.custom_sources is not None) and (entry['source'] not in self.custom_sources):
+                continue
+
+            # raise error if the entry does not contain the required metadata
+            self.contains_required_metadata(entry)
```
mruwnik, line 101: this means that a single faulty item will blow up the whole run - is that desired?

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+                signature = f"Title: {title}, Authors: {get_authors_str(authors)}"
+            else:
+                signature = f"Title: {title}"
+
+            # We use the text_splitter to get the chunks from the entry, 
+            # and we add a metadata element for each new chunk we add to the dataset.
+            chunks = text_splitter.split(text, signature)
+            num_chunks = len(chunks)
+            
+            for i in range(num_chunks):
```
mruwnik, line 131: how about `for i, chunk in enumerate(1, chunks):`?

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+    def show_stats(self):
+        print(f'Number of entries by source: {self.entries_per_source_count}')
+        print(f'Total entries count: {self.total_counts["entries"]}')
+        print(f'Total character count: {self.total_counts["chars"]}')
+        print(f'Total word count: {self.total_counts["words"]}')
+        print(f'Total sentence count: {self.total_counts["sentences"]}')
+        print(f'Total chunk count: {self.total_counts["chunks"]}')
 
     def get_embeddings(self):
         def get_embeddings_at_index(texts: str, batch_idx: int, batch_size: int = 200): # int, np.ndarray
```
mruwnik, line 160: is it possible to do the assignments in the for loop directly via numpy? I have a vauge impression that can massively speed things up. Though I might be thinking of something else...

src/dataset/create_dataset.py
```diff
@@ -269,15 +171,16 @@ def get_embeddings_at_index(texts: str, batch_idx: int, batch_size: int = 200):
...
         with concurrent.futures.ThreadPoolExecutor() as executor:
             futures = [executor.submit(
                 get_embeddings_at_index, 
-                self.embedding_strings[batch_idx:batch_idx+batch_size], 
+                embedding_strings[batch_idx:batch_idx+batch_size], 
                 batch_idx,
-                len(self.embedding_strings[batch_idx:batch_idx+batch_size])
-            ) for batch_idx in range(0, len(self.embedding_strings), batch_size)]
+                len(embedding_strings[batch_idx:batch_idx+batch_size])
+            ) for batch_idx in range(0, len(embedding_strings), batch_size)]
```
mruwnik, line 183: how about having a chunker function that receives a list (or even just an iterable) and splits it up into chunks of n items long? It would make this a lot clearer

src/dataset/create_dataset.py
```diff
@@ -269,15 +171,16 @@ def get_embeddings_at_index(texts: str, batch_idx: int, batch_size: int = 200):
         rate_limit = 3500 / 60  # Maximum embeddings per second
 
         start = time.time()
-        self.embeddings = np.zeros((len(self.embedding_strings), LEN_EMBEDDINGS))
+        embedding_strings = [chunk_data['text'] for chunk_data in self.metadata]
+        self.embeddings = np.zeros((len(embedding_strings), LEN_EMBEDDINGS))
 
         with concurrent.futures.ThreadPoolExecutor() as executor:
             futures = [executor.submit(
```
mruwnik, line 178: how about using [map](https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.map) here? e.g.
```
results = executor.map(get_embeddings_at_index, chunks)
for (batch_idx, embeddings) in results:
    (...)
```
This assumes that there is a `chunks` iterable with the embeddings already chunked up

src/dataset/create_dataset.py
```diff
@@ -289,113 +192,32 @@ def get_embeddings_at_index(texts: str, batch_idx: int, batch_size: int = 200):
...
-        authors = [authors_string.strip()]
-    return authors
+        authors_lst = authors_lst[:3]
+        authors_str = ", ".join(authors_lst[:-1]) + " and " + authors_lst[-1]
+    return authors_str
 
 def standardize_date(date_string, default_date='n/a'):
     try:
         dt = parse(date_string)
         return dt.strftime('%Y-%m-%d')
```
mruwnik, line 221: would `dt.date().isoformat()` work?

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+            self.total_counts['sentences'] += len(split_into_sentences(text))
+            self.total_counts['chunks'] += len(chunks)
+
+    def show_stats(self):
+        print(f'Number of entries by source: {self.entries_per_source_count}')
+        print(f'Total entries count: {self.total_counts["entries"]}')
+        print(f'Total character count: {self.total_counts["chars"]}')
+        print(f'Total word count: {self.total_counts["words"]}')
+        print(f'Total sentence count: {self.total_counts["sentences"]}')
+        print(f'Total chunk count: {self.total_counts["chunks"]}')
```
mruwnik, line 157: [NIT] would it be possible to change these `print`s to `logger.info`?

src/dataset/create_dataset.py
```diff
@@ -19,240 +20,141 @@
...
+            else:
+                signature = f"Title: {title}"
+
+            # We use the text_splitter to get the chunks from the entry, 
+            # and we add a metadata element for each new chunk we add to the dataset.
+            chunks = text_splitter.split(text, signature)
+            num_chunks = len(chunks)
+            
+            for i in range(num_chunks):
+                self.metadata.append({
```
mruwnik, line 132: this could be a potential problem on machines without much memory - it might be better to change this method into a a generator. Though that would mess up the `save_data` function...

src/main.py
```diff
@@ -34,130 +34,25 @@
...
-    
+    dataset.show_stats()
+
+def load_pkl_and_display_stuff():
+    with open(PATH_TO_DATASET_DICT_PKL, 'rb') as f:
+        dataset_dict = pickle.load(f)    
+    print(dataset_dict)
+
 @retry(wait=wait_random_exponential(min=1, max=20), stop=stop_after_attempt(4))
 def get_embedding(text: str) -> np.ndarray:
```
mruwnik, line 51: is this function used?

mruwnik (commented):
A PR that is mostly deletions - my favorite kind!

src/dataset/create_dataset.py
Thomas-Lemoine: alright. Same for summary?

mruwnik: yup. I just merged a PR that should have ensured that, and I updated the datasets on HF, so it should be ok. Let me know if I messed up :D

src/dataset/create_dataset.py
Thomas-Lemoine: Good idea. Since looking at the ARD code yesterday I also learned about staticmethod and should this not be a case where I call it like 
@staticmethod
def _validate_required_metadata, since I wouldn't want someone to use this function outside of the class since that wouldn't make much sense?

Thomas-Lemoine: and because it doesn't use self. *

mruwnik: yup, a staticmethod makes sense. It's not really an issue - I personally like to use them as I try to keep functions as pure as possible and that helps mark them are purer. I also prefer to use raw functions, rather than classes, but that's a different issue...
`_validate_required_metadata` makes sense, yes. Don't worry too much about naming things though. After all - it's one of the [two hard things in CS](https://twitter.com/secretGeek/status/7269997868?ref_src=twsrc%5Etfw%7Ctwcamp%5Etweetembed%7Ctwterm%5E7269997868%7Ctwgr%5E279ff3998e5bf3ccaa3d6723d084070236fa1cf3%7Ctwcon%5Es1_&ref_url=https%3A%2F%2Fwww.martinfowler.com%2Fbliki%2FTwoHardThings.html)

src/dataset/create_dataset.py
Thomas-Lemoine: my brother, who is currently creating a new feature from my branch, did something similar to this, makes sense

src/dataset/create_dataset.py
Thomas-Lemoine: I think so, because at this point we want to make sure the ARD has correct entries. If it doesn't, it indicates some of its code has to be changed asap to be standardized. Although I can use logging.error instead? Like would do something like
```
# raise error if the entry does not contain the required metadata
try:
    self._validate_required_metadata(entry)
except MissingDataException as mde:
    logging.error(str(mde))  # Log the error message
except Exception as e:
    raise e
```
Although, I don't have a logging thing yet, will have to look at ARD to see how they did it

src/dataset/create_dataset.py
Thomas-Lemoine: Yeah I didn't really touch get_embeddings, it's a bit confusing and complicated, but it can be improved for sure I'll test things later

src/dataset/create_dataset.py
Thomas-Lemoine: Yes, certainly, good idea

src/dataset/create_dataset.py
Thomas-Lemoine: Are you saying we essentially don't want to have to hold into memory the entire content of the dataset, which is what we do by constructing self.metadata? If so, yeah I agree, if we add the information from the entries into the sqlite3 database that removes the necessity of storing self.metadata at all, correct? 

Another thing this reminds me of is that I may not understand how the streaming on the load_dataset() function works, because my understanding is that, as we go through the iterable, we load more and more bits of the dataset, but at the end, we hold the entire dataset into memory, and maybe we crash if that's too big for our memory - so wouldn't we want to go through the dataset but forget about previously seen entries? Maybe that's what it does, I'm not sure

src/main.py
Thomas-Lemoine: The get_embedding ones isn't currently, it's just there to be used to test whether the openAI embeddings worked correctly. I can check whether dataset.embeddings[0] == get_embedding(dataset.metadata[0])

src/dataset/create_dataset.py
Thomas-Lemoine: I'm checking whether they're non-empty though, meaning a summary of [] triggers the error currently, which is common. I can instead check that 'key' in entry.keys(), which would check if 'summary' is a key of the entry even if it's an empty list or whatever

src/dataset/create_dataset.py
ccstan99: Unless I'm misunderstanding... Since you are already rate-limited to 3500 embeddings per minute, there may not be a benefit to have everything in memory all at once. Personally, I didn't run into memory problems running in the colab notebook and SentenceTransformer embeddings were able to batch it properly but OepnAI may be different. 

However, I do think we should still store most of the key metadata (id, source, title, authors, date_published, chunked_text but can skip full_text and summary for now) in pinecone as well. Sqlite3 should only be used to keep track of data for ingesting ARD, chunking, and embedding. Whatever the chatbot needs to query & generate answers should be stored in pinecone to reduce an extra dependencies (risk of failure) at runtime. The metadata feature of pinecone is supposed to be one of its benefits over other vector DBs.

ccstan99: Although... it might be nice to have a counter in sqlite to keep track of which entries in the ARD are used as sources most frequently. Anyway, once the database is online, there's a bunch of other things we would like to track. Not sure sqlite can handle the load but that's a separate discussion.

Thomas-Lemoine: I definitely agree we don't want everything in memory at once. It's good that it works for the colab but it's still practical if my brother and I can test it without those concerns, but also it's not top priority I guess as long as it works. And alright for pinecone, since lw has less entries than before, storing those metadata values should be OK. So I guess in that case the DB is only used for constructing the pinecone, or when we want to get information from keys that aren't in the 8 basic ones, like tags or "book_title" and so on?

src/dataset/create_dataset.py
```diff
@@ -74,7 +74,7 @@ def _validate_required_metadata(entry: Dict[str, Any]):
 
         # Check that the 8 primary metadata keys all have the correct type and the key exists
         for key, key_type in metadata_types.items():
-            if not entry[key]:
+            if not key in entry.keys():
```
mruwnik, line 77: is the `.keys()` needed here?

Thomas-Lemoine: no, fixed

mruwnik (approved):

Thomas-Lemoine:
looks like we're gonna scrap this PR since my brother's one is more complete

FraserLee:
> looks like we're gonna scrap this PR since my brother's one is more complete

Should we close the PR?

PR #39, opened 2023-06-23
Update sql and pinecone

henri123lemoine:


src/dataset/update_dataset.py
```diff
@@ -2,82 +2,56 @@
...
     def update_source(self, source: str):
-        self.logger.info(f"Updating {source} entries...")
+        logger.info(f"Updating {source} entries...")
 
-        iterable_data = load_dataset('StampyAI/alignment-research-dataset', source, split='train', streaming=True)
+        iterable_data = load_dataset(ARD_DATASET_NAME, source, split='train', streaming=True)
         iterable_data = iterable_data.map(self.preprocess)
         iterable_data = iterable_data.filter(lambda entry: entry is not None)
-        iterable_data = iterable_data.filter(lambda entry: self.db.upsert_entry(entry))
+        iterable_data = iterable_data.filter(lambda entry: self.sql_db.upsert_entry(entry))
```
mruwnik, line 39: [NIT] the usual approach is to chain these calls, like:
```
iterable_data = load_data_set(
   source, ...
).map(self.preprocess).filter(
   ...
)
```

src/main.py
```diff
@@ -1,30 +1,30 @@
...
-def update_database_and_pinecone():
+def update_sql_and_pinecone_dbs():
     updater = ARDUpdater(
         min_tokens_per_block=200,
         max_tokens_per_block=300,
     )
-    updater.update(['yudkowsky_blog'])
+    if RESET_DBS:
+        updater.reset_dbs()
+    updater.update(['gwern_blog'])
```
mruwnik, line 23: why this specifically? Is just for testing whether things work?

henri123lemoine: Yes, gwern's blogs are one of the shortest sources to test for, and I tend to alternate between the other small sources because running twice with the same source does nothing.

mruwnik: hmm. This doesn't sit right with me for some reason...

src/main.py
```diff
@@ -1,163 +1,31 @@
...
-    plt.savefig("bla.png")
+    if RESET_DBS:
+        updater.reset_dbs()
+    updater.update(['gwern_blog'])
 
-    
-    
 
+def main():
+    update_sql_and_pinecone_dbs()
```
mruwnik, line 27: why not just directly call this function, rather than going through this `main()` call?

mruwnik (dismissed):

src/main.py
henri123lemoine: Because I was told that it's more modular and respects norms/conventions, but I realize it doesn't matter in this case. I don't mind either way, and can change it back

src/dataset/pinecone_db_handler.py
```diff
@@ -17,10 +17,8 @@ def __init__(
     ):
         self.index_name = PINECONE_INDEX_NAME
         
-        PINECONE_API_KEY = os.getenv("PINECONE_API_KEY")
-        PINECONE_ENVIRONMENT = os.getenv("PINECONE_ENVIRONMENT")
-        assert PINECONE_API_KEY, "PINECONE_API_KEY environment variable not set."
-        assert PINECONE_ENVIRONMENT, "PINECONE_LOCATION environment variable not set."
+        PINECONE_API_KEY = os.environ["PINECONE_API_KEY"]
```
mruwnik, line 20: is this change explicitly for error handling? Why not move them to `settings.py`?

henri123lemoine: The `os.environ["PINECONE_API_KEY"]` instead of `os.getenv("PINECONE_ENVIRONMENT")` + assert was just a more concise version that did the ~same thing as before, where it calls an error if the env variable is not defined. And I just did, good idea.

src/dataset/pinecone_db_handler.py
```diff
@@ -0,0 +1,100 @@
...
+                            'authors': entry['authors'],
+                            'text': chunk,
+                        } for chunk in chunks
+                    ]
+                )
+            ),
+            batch_size=upsert_size
+        )
+    
+    def upsert_entries(self, entries_batch, chunks_batch, chunks_ids_batch, embeddings, upsert_size=100):
```
mruwnik, line 54: isn't this pretty much `upsert_entry` with an extra loop? i.e. it might be worth either calling `upsert_entry` from this function, or abstracting it away so that there is an additional function that does the actual work, which is called by these functions

mruwnik (approved):
LGTM, but I'm by no means an expert on ML stuff

src/dataset/pinecone_db_handler.py
henri123lemoine: Yes, but the reason it's faster is that it makes fewer index.upsert upsert calls, which used to be one of the major speed bottlenecks. It now calls index.upsert once per CHUNK_SIZE chunks instead of once per entry. Calling index.upsert with 10 vectors 100 times is much slower than calling it once with 1,000 vectors, which is the reason this whole batching endeavor was worth implementing in the first place. My (admittedly limitted) testing seems to show it's much faster than it used to be. It still takes over two hours to update the full dataset*, but when we'll run src/main.py just to update the current dataset with the past week's posts/papers/etc, it will only take a few seconds/minutes.
There are certainly ways to make the code cleaner, but I don't know of a simple way to make it cleaner while not making it slower. The process of creating the 'vectors' variable could be done in another method, I suppose.

*on colab with a GPU. Parenthetically, I tested it there because I wanted to use the free sentence transformer for testing and running it on my computer would have taken hundreds of hours. It also seems like Pinecone upserts are much faster on colab than on my computer. It's worth noting that the Pinecone documentation mentions this is normal and that google colab/google cloud/azure cloud/amazong cloud are much faster than home networks for pinecone upserts. I don't know if this means we'll essentially be forced to update the dataset on these cloud servers, but in any case I have a google colab that gets the most recent version of the right stampy-chat branch from the github repo and updates the pinecone dataset from there. One issue to fix however is to store the sql database externally, possibly on google drive, to not have to start from scratch every time.

mruwnik: No, you're totally correct on that front, I was thinking more a step earlier, eg.
```
def upsert(self, ids, embeddings, chunks, upsert_size):
    data = [
                        {
                            'entry_id': entry['id'],
                            'source': entry['source'],
                            'title': entry['title'],
                            'authors': entry['authors'],
                            'text': chunk,
                        } for chunk in chunks
    ] 
    return self.index.upsert(
        vectors=list(zip(ids, embeddings.tolist(), data)), batch_size=upsert_size
   )

def upsert_entry(self, entry, chunks, embeddings, upsert_size=100):
    self.upsert(
        [f"{entry['id']}_{str(i).zfill(6)}" for i in range(len(chunks))],
        embeddings,
        chunks
       upsert_size
   )

 def upsert_entries(self, entries_batch, chunks_batch, chunks_ids_batch, embeddings, upsert_size=100):
    self.upsert(chunks_ids_batch, embeddings, (chunk for entry in entries_batch for chunk in chunks_batch), upsert_size)

Issue #46, opened 2023-07-05
move some work to backend, improve algorithms


FraserLee:
Right now there's some text processing that happens in the front-end, including
- Unification of formats for single citations with Regices (the LLM isn't particularly consistent with its output)
- Unification of formats for citation groups
- Reordering citations within one message *(such that the first is* `[1]` *, the second* `[2]` *, etc)*
- Reordering citations such that each subsequent message starts counting at `n + 1`

Some (*all?*) of these should be moved to the backend in python. This would improve citation quality for alternate front-ends, including any ops retroactively looking at logs.

Additionally we're currently doing all this stuff through string manipulation when really it's better suited to some custom [ropes](https://en.wikipedia.org/wiki/Rope_(data_structure)). We could even consider serializing as ropes such that the front-end doesn't need to re-parse out where citations are, but my intuition here is that the added json overhead wouldn't be worth the trade-off.

If someone's particularly excited for this issue, I can drop some links to locations in the codebase to get started. If not, I'll tackle it at some point.

FraserLee:
After some investigation and a few prototypes, I believe the form of this I had initially envisioned is inviable. It would mean either sacrificing working citations while streaming, or refactoring to a diff-syncing protocol (likely hand-rolled - there doesn't seem to be much out of the box in this space).

Just as an example of why this is non-trivial, if the LLM outputs
```
chunk 1: "foo [a, b, c,"
chunk 2: " d, e] bar"
```
then there's no purely incremental message we can send without to both process out citations (`"foo [a, b] bar"` -> `"foo ", (citation object 1), (citation object 2), " bar"`) while sending our chunks in sync with the ones we recieve.

Possible solutions:
- wait until entire message received, don't stream in real time **->** terrible user experience
- re-send the entire message with every new chunk **->** quadratic order network usage, not an option
- come up with a scheme to send diffs instead of purely incremental messages **->** might work, but technically complicated and probably pretty fragile
- maintain a n chunk buffer such that we can always process far enough ahead before sending data **->** most feasible, will proceed with this option.

Issue #6, opened 2023-04-30
Testing


ccstan99:
- Develop a set of benchmark questions to evaluate effectiveness of prompts and parameters.
- Solicit editors and general user feedback to stress test conversational functionality

ccstan99:
Might be worthwhile to still have API functionality to automate testing & benchmarking quality of:
(1) semantic search retrieval and 
(2) also generated answer to standalone questions

ccstan99:
Ideally, it would be helpful to have a script that accepts a [google sheet](https://docs.google.com/spreadsheets/d/12sX5uKvkZZRTIdFfUZKpYMVN-UCf5Op-bI8rU_9kzw8/edit#gid=0) or text file as input with list of benchmark questions. The output are results on a separate [spreadsheet tab](https://docs.google.com/spreadsheets/d/12sX5uKvkZZRTIdFfUZKpYMVN-UCf5Op-bI8rU_9kzw8/edit#gid=385046677) or some format with a timestamp in the filename. The file should be easily imported into a spreadsheet with the following columns for each benchmark question:

- Query
- Generated Response
- Citations
- Prompt Template
- Source Block 1
- Source Block 2
- ... 
- Source Block N

If the output is CSV or TSV, make sure delimiters work and characters in text data are properly escaped. Essentially, we should be able to reconstruct the entire input to the LLM as well as its generated output. 

For now, let's focus on developing a solid prompt for standalone questions and assume chat history is empty. When we start debugging conversations, we'll need to rephrase the query based on the chat history before fetching context blocks and also log that.

This [spreadsheet](https://docs.google.com/spreadsheets/d/12sX5uKvkZZRTIdFfUZKpYMVN-UCf5Op-bI8rU_9kzw8/) with sample inputs and outputs can also be used as a central test "hub" where problematic questions can be added to the list of benchmark questions. New tabs can be added to track candidate prompts that are worth evaluating.

markovial:
Prompt : Can you explain the t-AGI framework?
Generated output: an explainer of t-AGI as tool AGI
Expected output : Distillation of [Clarifying and predicting AGI](https://www.lesswrong.com/posts/BoA3agdkAzL6HQtQP/clarifying-and-predicting-agi)

I tried another time after clearing the cache.

Prompt: what is t-AGI?
Generated output: an explainer of t-AGI as task directed AGI
[...]

PR #50, opened 2023-07-08
Prompt Eng

FraserLee:
# Prompt Engineering Workflow

As I had said previously, our prompt is a function - not a template. Certain parameters like what goes in a system vs a user prompt, how we delineate sources, where to break messages, how to handle history, etc - it's too much weight to be cleanly represented with anything but code. That said, this process should be fairly simple and straightforwards even to non-technical people:

1. Edit the function `construct_prompt` in `api/chat.py`
2. `cd api`
3. run `pipenv run python3 prompteng/prompteng.py`
4. open the google sheet
5. click `file->import`. Click `upload`. Drag and drop the file `api/prompteng/data/answers.csv`
6. press `cmd + a`. Click `format->wrapping->wrap`. Adjust column widths to preference.

## Possible future improvements

- Copying over CSV manually is burdensome, but my efforts to try to get google sheets api [didn't work out](https://discord.com/channels/677546901339504640/1125882946952384635/1127278264503779368). Setting this up backed by our own DB could be nice.

- We could throw GPT-4 in the loop to evaluate answers on some simple criteria (*eg "Does the following answer contain any information not in the sources? Write about three sentences of explanation, then on a new line answer either YES or NO"*).

- Although right now the prompt is loose enough that we need to keep it as a function, there might come a point where we nail down outer structure (eg 1 system message, 1 user message, system for sources in a certain way, final message of X, etc). We're not there yet, but when we reach this point we could have the instruction text parts of that stored in the spreadsheet and more easily modified and tracked.

- I initially had the output format closer to what was described in #6 with the sources in separate columns to the right of the data, but found that to be a lot less readable than having them in middle lines between the answers. Still not sure if there might be a better format possible.

- solving #46 would make the output contain less unused sources

Would love some feedback around this one before I merge!


FraserLee:
https://docs.google.com/spreadsheets/d/1kUb6Mtn1fXLd_iRHEOOnFjpyv0TA5I2Q-R1PtAG7ARw/edit?usp=sharing

api/chat.py
```diff
@@ -110,7 +114,13 @@ def construct_prompt(query: str, history: List[Dict[str, str]], context: List[Bl
...
     question_prompt = f"In your answer, please cite any claims you make back to each source " \
                     f"using the format: [a], [b], etc. If you use multiple sources to make a claim " \
-                    f"cite all of them. For example: \"AGI is concerning [c, d, e].\"\n\nQ: " + query
+                    f"cite all of them. For example: \"AGI is concerning [c, d, e].\"" \
+                    "\n\nMaintain a tone that is vehemently pro-humanity, and treat the" \
+                    "issue with the gravity of a potentially existential threat. Do not" \
+                    "be afraid to take strong sides, nor feel every claim need be hedged" \
+                    "with its inverse. If something is important or true, state it" \
+                    "unequivocally. Stress that it’s imperative that we act soon." \
+                    "\n\n\nQ: " +query
```
ccstan99, line 123: If you're going to update the prompt, can you please manually update the stampy-chat-test gsheet (https://docs.google.com/spreadsheets/d/12sX5uKvkZZRTIdFfUZKpYMVN-UCf5Op-bI8rU_9kzw8/edit#gid=385046677)? Add a tab with today's date and generated outputs for the benchmark questions so we can track how changes are affecting the results. I'm personally not convinced those wording are in the direction we necessarily want to head. But I'll defer if it produces desirable outcomes, as judged by Rob and distillation fellows. So it'll be even better if you can note examples where changing the prompt improved outputs.

ccstan99 (approved):

api/chat.py
ccstan99: We have a general dev meeting tomorrow. If you're able to get a list of generated outputs to review before, that'd be nice. But don't stress over it if not.

> Maintain a tone that is vehemently pro-humanity, and treat the issue with the gravity of a potentially existential threat. Do not be afraid to take strong sides, nor feel every claim need be hedged with its inverse. If something is important or true, state it unequivocally. Stress that it’s imperative that we act soon.

In theory, the ChatGPT's RLHF should have induced pro-humanity. The prompt above seem likely to induce hallucinations which might reduce our credibility. So having output for reference would be helpful.

ccstan99: Also nitpicky but for long strings, it might be easier to read & maintain using triple quotes, i.e.
```
question_prompt = """In your answer, 
please cite any claims you make back to each source using the format: 
[a], [b], etc. If you use multiple sources to make a claim 
cite all of them. For example: 'AGI is concerning [c, d, e].'

Maintain a tone that is vehemently pro-humanity and treat the
issue with the gravity of a potentially existential threat. Do not
be afraid to take strong sides, nor feel every claim need be hedged
with its inverse. If something is important or true, state it
unequivocally. Stress that it’s imperative that we act soon.

Q: """ + query
```

FraserLee: Thanks for the note on the triple quote strings, I had forgotten python had those.

As for the prompt change, I really was just looking to test if the prompt eng tooling was useful. It seemed to have a positive effect on a few of the test questions so I left it in the PR. Will role it back now.

FraserLee: Ohh I remember why I wasn't using triple quoted strings. It renders all newlines and indentation into the string itself - not viable.
Screenshot 2023-07-09 at 2 43 25 PM

cvarrichio:
Just so that we don't duplicate, I'm nearly done with https://github.com/StampyAI/stampy-chat/issues/6.  I should push something within a few days.

ccstan99:
@cvarrichio This PR mostly sets up a means to easily benchmarked / test new prompts and generates the output to a CSV file. Frazer's looking for feedback on it. If you're able to play around with this new setup, it might speed up your own testing and refinement process.

cvarrichio:
I'm pretty sure this is exactly what I've been working on.  Pausing for now.

ccstan99:
@cvarrichio Shoot, you're right! I remember that you wanted to work on prompt engineering and now I remember mentioning that having the testing in place would make that work faster. I see that I even wrote it down on our planning docs, but Frazer wasn't as the meeting so probably didn't catch that. So sorry for the confusion. 

FYI, I added some specific things to try for prompt engineering #3, which hopefully be much easier to try with these changes.

FraserLee:
Ah sorry for stepping on your toes there, I thought you were more working on #2. If you have any improvements to my version feel free to open a second PR on top!

PR #57, opened 2023-07-11
create level modes

### LINKED ISSUE #33: Add modes and options
Adjust prompt for novice user (less jargon), editor/research (more technical details), skeptic (arguments, debate)

Additional comment from McGill Team's Plan:
Determine which improve results most with the next point) HyDE, HyQE, Debate, Comment, Synthesis, model used (gpt3.5, 4, claude, finetuned model, etc), “creative, balanced, or precise” a la Bing, temperature (could be best at 0, or it might make the chatbot less bold/ more boring. Experiment! Could be fought with specific prompts that encourage boldness for example). Problem of various ones is slow speed, but may lower hallucination, etc.

Fun chatbot! We want it to be more entertaining. Chatbot makes jokes, is bold in its answers, etc.

FraserLee:
closes #33 .

Let's keep the discussion on this PR just on the technical implementation. I expect we'll want to discuss what exact modes to implement synchronously in a live meeting since that relates more to project direction - better to have a full speed back and forth \:)

Screenshot 2023-07-10 at 9 58 48 PM

### DIFF
--- a/api/chat.py
+++ b/api/chat.py
@@ -48,7 +48,7 @@ def cap(text: str, max_tokens: int) -> str:
 
 
 
-def construct_prompt(query: str, history: List[Dict[str, str]], context: List[Block]) -> List[Dict[str, str]]:
+def construct_prompt(query: str, mode: str, history: List[Dict[str, str]], context: List[Block]) -> List[Dict[str, str]]:
 
     prompt = []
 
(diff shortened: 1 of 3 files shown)


henri123lemoine (changes requested):
```py
    question_prompt = "In your answer, please cite any claims you make back to each source " \
                      "using the format: [a], [b], etc. If you use multiple sources to make a claim " \
                      "cite all of them. For example: \"AGI is concerning [c, d, e].\n\n"
```
There should be a closing quotation mark after '\"AGI is concerning [c, d, e]'.
So,
```
"cite all of them. For example: \"AGI is concerning [c, d, e].\n\n"
```
->
```
"cite all of them. For example: \"AGI is concerning [c, d, e].\"\n\n"
```
Other than that, it all looks good.

Thomas-Lemoine:
To what henri said, I'm guessing you still need the \ before each quote characters

Thomas-Lemoine (commented):
This is a nice addition imo. I personally kind of prefer "concise" to "crux" because crux is more specific and so for some questions or circumstances it might not make sense to have a "crux" response, maybe? Thoughts?

FraserLee:
Nice, thanks for the reviews

PR #59, opened 2023-07-12
add one particularly good question to the standard set

FraserLee:


FraserLee:
also sorted list in source

PR #65, opened 2023-07-20
Improve text splitter

henri123lemoine:


henri123lemoine:
I'll move this over to [alignment-research-dataset](https://github.com/StampyAI/alignment-research-dataset) asap and switch sql to mysql and then update the pinecone dataset once and for all.

Thomas-Lemoine (approved):
again, not sure about the name ARDUpdater but afaiu it won't matter since you're deleting it

PR #71, opened 2023-07-30
internal ai.safety.info links handled correctly 

### LINKED ISSUE #66: Internal https://aisafety.info links are handled incorrectly
Internal links are given currently as `/?state=XXXX`. We should run a pre-process step when displaying human authored answers to insert the correct base url.

Screenshot 2023-07-23 at 3 31 37 PM

FraserLee:
closes #66

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -1,8 +1,9 @@
 from flask import Flask, jsonify, request, Response
 from flask_cors import CORS, cross_origin
+from urllib.parse import unquote
 import dataclasses
 import json
-from urllib.parse import unquote
+import re
 
 from env import PINECONE_INDEX, log
 from get_blocks import get_top_k_blocks
@@ -65,7 +66,13 @@ def human(id):
     r = requests.get(f"https://aisafety.info/questions/{id}")
     log(f"clicked followup '{json.loads(r.text)['data']['title']}': https://stampy.ai/?state={id}")
 
-    return Response(r.text, mimetype='application/json')
+    # run a regex to replace all relative links with absolute links. Just doing
+    # a regex for now since we really don't need to parse everything out then
+    # re-serialize it for something this simple.
+    # <a href=\"/?state=6207\"> -> <a href=\"https://stampy.ai/?state=6207\">
+    text = re.sub(r'<a href=\\"/\?state=(\d+)\\">', r'<a href=\"https://stampy.ai/?state=\1\">', r.text)
+    
+    return Response(text, mimetype='application/json')
 
 # ------------------------------------------------------------------------------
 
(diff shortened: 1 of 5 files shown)


api/main.py
```diff
@@ -65,7 +66,13 @@ def human(id):
     r = requests.get(f"https://aisafety.info/questions/{id}")
     log(f"clicked followup '{json.loads(r.text)['data']['title']}': https://stampy.ai/?state={id}")
 
-    return Response(r.text, mimetype='application/json')
+    # run a regex to replace all relative links with absolute links. Just doing
+    # a regex for now since we really don't need to parse everything out then
+    # re-serialize it for something this simple.
+    # <a href=\"/?state=6207\"> -> <a href=\"https://stampy.ai/?state=6207\">
+    text = re.sub(r'<a href=\\"/\?state=(\d+)\\">', r'<a href=\"https://stampy.ai/?state=\1\">', r.text)
```
ccstan99, line 73: Probably better to use "https://aisafety.info/?state=" instead of stampy.ai

web/src/pages/index.tsx
```diff
@@ -75,7 +75,7 @@ const ShowCitation: React.FC<{citation: Citation, i: number}> = ({citation, i})
 
     // if we don't have a url, link to a duckduckgo search for the title instead
     const url = citation.url && citation.url !== ""
-                ? citation.url 
+                ? citation.url
                 : `https://duckduckgo.com/?q=${encodeURIComponent(citation.title)}`;
```
ccstan99, line 79: Nice! I didn't notice the duckduckgo search before. Hopefully won't be needed as much after ARD is cleaned up. One thing we discussed that would be nice (eventually) is to check if the citation.url is valid (returns content without errors) and give an alternate result if the it's down or removed for some reason

ccstan99 (approved):
This looks good to me. Tagging @mruwnik in case he has any suggestions.

Also, after you merge this and it gets redeployed, double-check if the /semantic endpoint still works. TJ had to restart the service and thought there may be something amiss in our workflow.

mruwnik (approved):

web/src/pages/index.tsx
FraserLee: Absolutely. It's really meant as more of a stopgap, but should be nice as a fallback if stuff still slips through the cracks.

On the topic, at some point we might want to have some system in place for flagging "removed" sources, since them being "down" can mean anything from a literal 404, or temporarily down site page, or non-response, to just an internal redirect, or private page. Not really any good automated way to detect the later camp.

api/main.py
FraserLee: resolved.

PR #72, opened 2023-08-09
also fall back to duckduckgo on in-text citations

FraserLee:
accidentally missed this one the first time around


ccstan99 (approved):

ccstan99 (approved):
Snuck in a commit to clean up the README

FraserLee:
> Snuck in a commit to clean up the README

Very fair, good changes. Still in the future it might be better to keep PRs single-purpose where possible, so the git history is more meaningful when using tooling.

PR #73, opened 2023-08-12
Inject glossary into generated answers

### LINKED ISSUE #69: Inject glossary into generated answers
Similar to the existing answers on stampy that utilize the glossary system, we should inject the glossary from [coda](https://coda.io/d/AI-Safety-Info_dfau7sl2hmG/Glossary_suZVl#_luIKu) or [gdocs](https://docs.google.com/document/d/1Rm7tN5zTErZq0E4qqAZoNVzClS94Kb_rQOLaViu2p6c/edit) into the generated answers.

This should save a whole bunch on tokens, because people would more often not need to ask a subsequent questions which then generate a couple of paragraphs worth of text.

[Here is an example](https://aisafety.info/?state=0_6953r6201r) of the current functioning glossary.

FraserLee:
Closes #69

### DIFF
--- a/web/src/glossary.tsx
+++ b/web/src/glossary.tsx
@@ -41,14 +41,21 @@ export const GlossarySpan: React.FC<{content: string}> = ({content}) => {
         const hover_content = item.contents;
         const pageid = item.pageid;
 
-        return `
-            <div class="glossary-hover" nowrap>${hover_content}</div>
-            <a href="https://aisafety.info/?state=${pageid}"
-               target="_blank"
-               class="glossary-link">
-                 ${match}
-            </a>
-        `;
+        if (pageid == undefined || pageid.trim() == "") {
+            return `
+                <div class="glossary-hover" nowrap>${hover_content}</div>
+                <span class="glossary-link">${match}</span>
+            `;
+        } else {
+            return `
+                <div class="glossary-hover" nowrap>${hover_content}</div>
+                <a href="https://aisafety.info/?state=${pageid}"
+                   target="_blank"
+                   class="glossary-link">
+                     ${match}
+                </a>
+            `;
+        }
 
     })}} />;
 }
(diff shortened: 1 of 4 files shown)


FraserLee:
One thing of note: right now the glossary is tiny (only 23 terms, many duplicates). I'm not certain where existing efforts are at to grow it, but it feels like we could add *a lot* of value for relatively little effort by expanding this. We could even try things like auto-generating glossary entries for every existing stampy article where it could make sense using gpt-4, then manually reviewing and correcting them. Minuscule effort, potentially a pretty good end result.

Current keywords: `agi, ai alignment, artificial general intelligence, chain of thought prompting, chain-of-thought, embedded agency, embedded agent, existential risk, goal misgeneralization, goodhart's law, instrumental convergence, instrumentally convergent goals, interpretability, large language model, llm, ood, optimizer, oracle, orthogonality thesis, out of distribution, terminal goal, terminal goals, "the big g,"`

web/src/glossary.tsx
```diff
@@ -0,0 +1,54 @@
...
+    }
+
+    const glossary = g.g;
+    const glossaryRegex = g.r;
+
+    // Otherwise, replace glossary terms with links. We can do this in
+    // O(n * sum of term lengths) by finding String.prototype.indexOf of
+    // each term in the glossary (since that'd probably be backed by KMP)
+    // but I think it should be faster to compile a regex state machine
+    // once and use that instead.
```
mruwnik, line 34: what about substrings? i.e. say you have 'good', 'goodness', 'goodhart's law', 'not very good' and 'blagoodblabla' as glossary entries. How would it handle the following phrases:

'this is good'
'the greater good is always the best'
'make sure goodharts law doesn't bite you'
'it's not very good if the word good get's inserted twice here'
'goodness me! Which definition of good should have been used here?'
'ritegoody is not how you describe how well someone can spell'

It might suffice to just sort the phrases by length before creating them, but you might want to check.

FraserLee: Tested: word breaks and pre-sorting do work.

web/src/pages/_app.tsx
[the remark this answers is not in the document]
FraserLee: I'm going to mark this as resolved for now, since the actual functionality of this PR is done. Will leave fetching the live glossary as another issue.

PR #86, opened 2023-08-23
Log queries

mruwnik:
I did some refactoring to prepare for logging queries etc. to mysql


web/src/settings.ts
```diff
@@ -0,0 +1,2 @@
+export const API_URL = process.env.NEXT_PUBLIC_API_URL || "http://127.0.0.1:3001";
+export const STAMPY_URL = process.env.STAMPY_URL || "http://127.0.0.1:8787"
```
mruwnik, line 2: this will have to be set to 'https://aisafety.info' wherever the deployed thing gets things from. CORS should work now, but only for that domain. It might be worth changing the UI to also accept localhost...

api/setup.py
FraserLee, line 1: What's the purpose of this file? Is it now the case that  the `install_requires` list and `api/Pipfile` have duplicate state that we manually need to keep in sync?

api/main.py
```diff
@@ -55,28 +53,7 @@ def chat():
...
-    r = requests.get(f"https://aisafety.info/questions/{id}")
-    log(f"clicked followup '{json.loads(r.text)['data']['title']}': https://stampy.ai/?state={id}")
-
-    # run a regex to replace all relative links with absolute links. Just doing
-    # a regex for now since we really don't need to parse everything out then
-    # re-serialize it for something this simple.
-    # <a href=\"/?state=6207&question=What%20is%20%22superintelligence%22%3F\">
-    #                               ⬇️
-    # <a href=\"https://stampy.ai/?state=6207&question=What%20is%20%22superintelligence%22%3F\">
-    text = re.sub(r'<a href=\\"/\?state=(\d+.*)\\">', r'<a href=\"https://aisafety.info/?state=\1\\">', r.text)
```
FraserLee, line 75: Good work attempting to delete the proxy function here, but as it stands there's two issues blocking merges
1. The code to patch up link addresses is deleted without replacement. You will need either:
    - reimplement URL swapping on the front-end, or
   -  change redirection rules such that `chat.aisaftey.info/?state=X` points to the same thing as `aisafety.info/?state=X`

2. This change means we're no longer logging which followups are clicked - info that it's absolutely critical that we log. If you're intent on moving things over, you'll need to add a separate call to log stuff.

api/src/stampy_chat/chat.py
```diff
@@ -141,7 +141,30 @@ def construct_prompt(query: str, mode: str, history: List[Dict[str, str]], conte
 import time
 import json
 
-def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str, str]], k: int = STANDARD_K, log: Callable = print):
+
+def check_openai_moderation(prompt: Prompt, query: str, log: Callable):
+    # 3. Run both the standalone query and the full prompt through
+    # moderation to see if it will be accepted by OpenAI's api
```
FraserLee, line 147: duplicate comment doesn't make sense when factoring this out of the function, probably accidental.
```suggestion
```

api/src/stampy_chat/chat.py
```diff
@@ -141,7 +141,30 @@ def construct_prompt(query: str, mode: str, history: List[Dict[str, str]], conte
...
+def remaining_tokens(prompt: Prompt):
+    # Count number of tokens left for completion (-50 for a buffer)
+    used_tokens = sum([
+        len(ENCODER.encode(message["content"]) + ENCODER.encode(message["role"]))
+        for message in prompt
+    ])
+    return NUM_TOKENS - used_tokens - TOKENS_BUFFER
+
+
+def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str, str]], k: int = STANDARD_K, log: Callable = logger.info):
```
FraserLee, line 167:
```python
log: Callable = logger.info
```
There's a few places across the codebase following this pattern. I understand why I wrote it that way when the fallback logger was `print`, but with your new logging changes does it still make sense to pass in a log function with a default parameter? If so, why?

api/src/stampy_chat/logging.py
```diff
@@ -0,0 +1,29 @@
...
+    def is_debug(self):
+        return self.isEnabledFor(DEBUG)
+
+    def query(self, query):
+        self.info('query: %s', query)
+
+    def response(self, response):
+        self.info('response: %s', response)
+
+    def moderation_issue(self, query, prompt_string, mod_res):
```
FraserLee, line 17: This function is incredible single purpose, I would recommend inlining into `check_openai_moderation` to simplify logic and readability.

web/src/glossary.tsx
FraserLee, line 1: Fine to leave this in right now, but in the future we should really have small isolated changes like this as their own tiny PRs. Easier to review, better git history, etc.

web/src/pages/index.tsx
```diff
@@ -435,7 +434,7 @@ const Home: NextPage = () => {
     // ----------------- HUMAN AUTHORED CONTENT RETRIEVAL ------------------
       const query_id = query.split("\n", 2)[0];
 
-      const res = await fetch(API_URL + "/human/" + query_id, {
+      const res = await fetch(`${STAMPY_URL}/questions/${query_id}`, {
```
FraserLee, line 437: see comments in `api/main.py`

api/src/stampy_chat/dataset_dl.py
FraserLee, line 1: Just to check: What's the purpose of moving everything in `api` to `api/src`?

FraserLee (changes requested):

FraserLee:
Lot of really great changes here, with just a few outstanding points I would either like addressed or like some more clarity on. Solid work!

henri123lemoine:
Shouldn't api/main.py be at location api/src/main.py? I don't get how to call main, since otherwise there are import issues.

api/setup.py
mruwnik: yes. Unfortunately. Pipfile is just a list of requirements, while setup.py has general metadata for the project

api/src/stampy_chat/chat.py
mruwnik: no. I left it so there were as few changes as possible - the next PR removes them

api/main.py
mruwnik: yeah, I'll remove this stuff from this PR

api/setup.py
FraserLee: Why do we need `setup.py`? I'm not familiar with this environment scheme.

api/src/stampy_chat/chat.py
FraserLee: Sounds good, though in the future it would be better to do changes like that on your first PR then pull them into the second.

api/src/stampy_chat/dataset_dl.py
FraserLee: Could you change things to do that in a separate PR?

mruwnik:
Closing this, as other PRs do the same

PR #88, opened 2023-08-24
Tidy up frontend

mruwnik:
This is mainly moving things around so the code is more modular. 
The search functionality has been refactored - there are comments showing what has changed


web/src/pages/index.tsx
```diff
@@ -271,306 +53,91 @@ const Home: NextPage = () => {
...
   useEffect(() => {
-    const mode = localStorage.getItem("chat_mode") as Mode || "default";
+    const mode = (localStorage.getItem("chat_mode") as Mode) || "default";
     setMode([mode, true]);
   }, []);
 
+  const updateCurrent = (current: CurrentSearch) => {
+    setCurrent(current);
+    if (current?.phase === "streaming") {
+      scroll30();
```
mruwnik, line 63: this might not be needed, but I wanted to keep as close to the original code as possible

web/src/pages/index.tsx
```diff
@@ -271,306 +53,91 @@ const Home: NextPage = () => {
...
+    const { result, followups } = await runSearch(
+      query,
+      query_source,
+      mode[0],
+      runningIndex,
+      entries,
+      updateCurrent
+    );
 
-    // ----------------------------- LLM BASED -----------------------------
```
mruwnik, line 297: this functionality is over in `useSearch.ts`

web/src/pages/index.tsx
```diff
@@ -271,306 +53,91 @@ const Home: NextPage = () => {
...
-      enable((f_old: Followup[]) => {
-        const f_old_filtered = f_old.filter((f) => f.pageid !== data.pageid && !fpids.has(f.pageid));
-        return [...f_new, ...f_old_filtered].slice(0, MAX_FOLLOWUPS); // this is correct, it's N and not N-1 in javascript fsr
-      });
-
-      scroll30();
+      setRunningIndex(runningIndex + ProcessText(result.content, 0)[1].size);
     }
+    setEntries((prev) => [...prev, result]);
+    enable(followups || []);
```
mruwnik, line 95: followups can be either a list of followups, or a function to update the current value, so both flows will work

FraserLee:
I'll defer on reviewing this till the others are merged since it depends on those changes, but broadly it seems like a lot of what this PR is doing is just increasing file count - doesn't inherently improve code quality. I'll check out the details after.

mruwnik:
This now points directly to main, so is independent of the other branches.

It does increase the count, in that a large part of it is moving components into their own files. 

The main changes are in the bit which actually queries the LLM/aisafety.info, which got split up into smaller (hopefully more understandable) functions. It took me a while to realise the the massive loop was simply a state machine. I changed it to only mutate a single state variable (the current item), and only append the exchange once it has finished.

I can split this up into smaller ones, but am a lot less certain that I'd do it properly, so would prefer not too. I will if requested, though

mruwnik:
I split it up into smaller ones, so will close this

PR #89, opened 2023-09-01
Move API files to a src folder

mruwnik:
Moves the API code into a src folder, to keep things nice and tidy. Also to avoid accidental name collisions etc.


api/chat.py
```diff
@@ -1,246 +0,0 @@
-# ------------------------------- env, constants -------------------------------
-from followups import multisearch_authored
-from get_blocks import get_top_k_blocks, Block
```
mruwnik, line 3: these lines were changed after this was moved, so git thinks this was deleted, rather than moved

api/src/stampy_chat/chat.py
```diff
@@ -0,0 +1,247 @@
+# ------------------------------- env, constants -------------------------------
+from dataclasses import asdict
+from typing import List, Dict, Callable
+import openai
+import re
+import tiktoken
+import time
+
+from stampy_chat.followups import multisearch_authored
+from stampy_chat.get_blocks import get_top_k_blocks, Block
```
mruwnik, line 10: these two lines are the only change (prepended `stampy_chat.` and moved them lower, as local imports should be last)

api/src/stampy_chat/env.py
```diff
@@ -15,6 +15,8 @@
 LOGGING_URL      = os.environ.get('LOGGING_URL')
 PINECONE_INDEX   = None
 
+FLASK_PORT = int(os.environ.get('FLASK_PORT', '3001'))
```
mruwnik, line 18: otherwise there's a port conflict, as both the frontend and backend use the same port

web/src/pages/index.tsx
```diff
@@ -1,5 +1,3 @@
-const API_URL = process.env.NEXT_PUBLIC_API_URL || "http://127.0.0.1:3000";
```
mruwnik, line 1: this is small enough and I keep having to update it, so I included it here

henri123lemoine (dismissed):
I very strongly approve these changes, though wouldn't it make more sense to split this into two PRs, one that moves the files and one that makes the additional changes? Would that fix the git weirdness? In any case, this change makes much more structural sense

henri123lemoine:
Is it me or the chat.py file still exists in api, but that it's just empty? Or am I pulling incorrectly

mruwnik:
nope, it's not you. Try `git checkout separate-src && git reset --hard HEAD~4 && git pull`

henri123lemoine:
So, right now the structure is this:
```
.
├── Pipfile
├── Pipfile.lock
├── main.py
├── nixpacks.toml
├── prompteng
│   ├── data
│   │   ├── answers.csv
│   │   └── questions.csv
│   ├── prompteng.py
│   └── readme.md
├── setup.py
└── src
    ├── Stampy_chat.egg-info
    │   ├── PKG-INFO
    │   ├── SOURCES.txt
    │   ├── dependency_links.txt
    │   ├── requires.txt
    │   └── top_level.txt
    └── stampy_chat
        ├── __init__.py
        ├── chat.py
        ├── dataset_dl.py
        ├── env.py
        ├── followups.py
        └── get_blocks.py
```
And in `prompteng/prompteng.py`, there's the following code:
```py
from pathlib import Path
import sys

sys.path = [str(Path(__file__).parent.parent)] + sys.path
from env import PINECONE_INDEX
from chat import talk_to_robot_simple, set_debug_print
```
My IDE is complaining at me. Wouldn't it make more sense to either import like this,
```py
from ..src.stampy_chat.env import PINECONE_INDEX
from ..src.stampy_chat.chat import talk_to_robot_simple, set_debug_print
```
, or to restructure the code by e.g. moving prompteng in stampy_chat or something like that to make the import nicer?

mruwnik:
good catch - it should be better now

henri123lemoine:
I don't get why this would work. My tests seem to show that before the `sys.path = [str(Path(__file__).parent.parent)] + sys.path`  line, `sys.path` is "~/stampy-chat/api/prompteng", and after that line, `sys.path` is "~/stampy-chat/api". So right after when you import the index using `from stampy_chat.env import PINECONE_INDEX`, how does it find it, when the index is only in src.stampy_chat.env from `stampy-chat/api`'s perspective? In any case, any reason why you think (1) is better than (2)?
(1)
```py
sys.path = [str(Path(__file__).parent.parent)] + sys.path
from stampy_chat.env import PINECONE_INDEX
from stampy_chat.chat import talk_to_robot_simple, set_debug_print
```
(2)
```py
from ..src.stampy_chat.env import PINECONE_INDEX
from ..src.stampy_chat.chat import talk_to_robot_simple, set_debug_print
```

mruwnik:
sys.path is a list, at least it should be, so this should just prepend the parent folder. Which if you run it as `pipenv run python prompteng/prompteng.py` shouldn't be needed? I'm guessing this was to get round the python imports being from the parent folder. Which is why I prefer proper, fully qualified imports - you don't have to do this :D

I actually prefer option 3, i.e. just
```
from stampy_chat.env import PINECONE_INDEX
from stampy_chat.chat import talk_to_robot_simple, set_debug_print
```
without the path magic, but wanted to change as little as possible here

henri123lemoine:
Ahhh I understand now, that makes sense. Thanks

henri123lemoine (approved):

henri123lemoine:
(though there are still some IDE warnings from `from stampy_chat.env import PINECONE_INDEX` and `from stampy_chat.chat import talk_to_robot_simple, set_debug_print`)

mruwnik:
I'll get round to those later :D

PR #90, opened 2023-09-01
MySQL support

mruwnik:
This is just the MySQL handling stuff, no actual usage

### DIFF
--- a/api/migrations/alembic.ini
+++ b/api/migrations/alembic.ini
@@ -1,0 +1,110 @@
+# A generic, single database configuration.
+
+[alembic]
+# path to migration scripts
+script_location = migrations
+
+# template used to generate migration file names; The default value is %%(rev)s_%%(slug)s
+# Uncomment the line below if you want the files to be prepended with date and time
+# see https://alembic.sqlalchemy.org/en/latest/tutorial.html#editing-the-ini-file
+# for all available tokens
+# file_template = %%(year)d_%%(month).2d_%%(day).2d_%%(hour).2d%%(minute).2d-%%(rev)s_%%(slug)s
+
+# sys.path path, will be prepended to sys.path if present.
+# defaults to the current working directory.
+prepend_sys_path = .
+
+# timezone to use when rendering the date within the migration file
+# as well as the filename.
+# If specified, requires the python-dateutil library that can be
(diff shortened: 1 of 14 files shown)


api/src/stampy_chat/db/session.py
```diff
@@ -0,0 +1,82 @@
...
+@contextmanager
+def make_session(auto_commit=False):
+    with Session(engine, autoflush=False) as session:
+        yield session
+        if auto_commit:
+            session.commit()
+
+
+class ItemAdder:
+    """A helper class to manage adding and flushing items to the database.
```
mruwnik, line 25: this isn't needed here, but I couldn't be bothered to mess about with deleting it and inserting it elsewhere

api/src/stampy_chat/db/models.py
```diff
@@ -0,0 +1,92 @@
...
+    query: Mapped[str] = mapped_column(String(1028))
+
+    # The full prompt as sent to the LLM
+    prompt: Mapped[Optional[str]] = mapped_column(LONGTEXT)
+
+    # Whatever the LLM returns
+    response: Mapped[Optional[str]] = mapped_column(LONGTEXT)
+
+    # The ids of the chunks used for the prompt
+    chunks: Mapped[Optional[str]] = mapped_column(String(1028))  # TODO: Change this to a proper format
```
mruwnik, line 63: the ids are like 10 chars each, and there shouldn't be that many of them, so 1028 bytes should suffice?

local_db.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env bash
```
mruwnik, line 1: this is exactly the same as in the ARD repo, other than it using a different database. The nice thing about this is that there will only be one MySQL Docker container hanging around, so it should be easy to check things out from the same mysql session

api/migrations/alembic.ini
```diff
@@ -0,0 +1,110 @@
+# A generic, single database configuration.
```
mruwnik, line 1: boilerplate stuff

mruwnik:
ping

mruwnik:
ping

ccstan99 (approved):

PR #92, opened 2023-09-01
Log to mysql

mruwnik:
Adds a logger that:
* handles all normal logging thingies
* is controllable via the `LOG_LEVEL` env variable
* will send all messages that it receives on to Discord if the message is from a `stampy_chat` module and it has a level at least at `DISCORD_LOG_LEVEL`
* will create appropriate MySQL entries for each interaction

### DIFF
--- a/api/src/stampy_chat/logging.py
+++ b/api/src/stampy_chat/logging.py
@@ -1,0 +1,75 @@
+import json
+from logging import *
+from discord_webhook import DiscordWebhook
+from stampy_chat.env import LOG_LEVEL, DISCORD_LOG_LEVEL, DISCORD_LOGGING_URL
+from stampy_chat.db.session import ItemAdder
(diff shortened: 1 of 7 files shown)


mruwnik:
ping

api/src/stampy_chat/logging.py
```diff
@@ -0,0 +1,75 @@
...
+            return
+
+        self.to_discord(self.format(record))
+
+    def to_discord(self, message):
+        if not DISCORD_LOGGING_URL:
+            return
+
+        while len(message) > 2000 - 8:
+            m_section, message = message[:2000 - 8], message[2000 - 8:]
```
Aprillion, line 26: would some variable name for `2000 - 8` make sense? `discord_max_len_without_buffer`?

Aprillion (approved):

PR #98, opened 2023-09-09
Deduplicate blocks

ccstan99:
- Deduplicate by blocks by hash_id instead of id. Might need to rework if we really want to keep & track individual block ids but this is a quick fix to the duplicate source problem.
- Switch from railway to chat.stampy.ai:PORT when pinecone key is not specified.

### DIFF
--- a/api/src/stampy_chat/get_blocks.py
+++ b/api/src/stampy_chat/get_blocks.py
@@ -60,9 +60,9 @@ def get_top_k_blocks(index, user_query: str, k: int) -> List[Block]:
 
     if index is None:
 
-        print('Pinecone index not found, performing semantic search on alignmentsearch-api.up.railway.app endpoint.')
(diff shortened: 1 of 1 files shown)


api/src/stampy_chat/get_blocks.py
```diff
@@ -132,7 +132,7 @@ def get_top_k_blocks(index, user_query: str, k: int) -> List[Block]:
         group = list(group)
         if len(group) == 0: continue
 
-        group = group[:3] # limit to a max of 3 blocks from any one source
+        # group = group[:3] # limit to a max of 3 blocks from any one source
```
Thomas-Lemoine, line 135: @FraserLee do you remember why you set this limit at the time? I also agree to not having the limit

Thomas-Lemoine (approved):

Issue #55, opened 2023-07-10
Investigate usage limits


mruwnik:
Find out how many requests per second can be handled by the current system. This applies both to the server infrastructure, but also to the underlying LLM system

ccstan99:
[OpenAI LLM chat rate limits](https://platform.openai.com/docs/guides/rate-limits/overview)
Current defaults as of July 2023: 
3,500 RPM (requests per minute) and 90,000 TPM (tokens per minute)

henri123lemoine:
We can apply to increase our rate limits. Should we?
It's also important to note that, should we decide to use it, GPT-4's current rate-limit is [200](https://platform.openai.com/account/rate-limits). OpenAI will increase that number over time.

henri123lemoine:
Current rate limits are of 200 messages or 40,000 tokens per minute, which will likely be reached in 5-6 questions if we use the full context window for every chat. That might be a problem.
We need to set up a system that handles GPT-4 rate limits by switching to ChatGPT if they occur. Additionally, we could limit spamming by slowing down the streaming if a single user sends queries at an impossible speed.

ishaan-jaff:
@mruwnik @ccstan99 @henri123lemoine 

i'm the maintainer of LiteLLM we allow you to **maximize your throughput/increase rate limits** -  load balance between multiple deployments (Azure, OpenAI)
**I believe litellm can be helpful here - and i'd love your feedback if we're missing something** 

Here's how to use it
Docs: https://docs.litellm.ai/docs/routing

```python
from litellm import Router

model_list = [{ # list of model deployments 
    "model_name": "gpt-3.5-turbo", # model alias 
    "litellm_params": { # params for litellm completion/embedding call 
        "model": "azure/chatgpt-v-2", # actual model name
        "api_key": os.getenv("AZURE_API_KEY"),
        "api_version": os.getenv("AZURE_API_VERSION"),
        "api_base": os.getenv("AZURE_API_BASE")
    }
}, {
    "model_name": "gpt-3.5-turbo", 
    "litellm_params": { # params for litellm completion/embedding call 
        "model": "azure/chatgpt-functioncalling", 
        "api_key": os.getenv("AZURE_API_KEY"),
        "api_version": os.getenv("AZURE_API_VERSION"),
        "api_base": os.getenv("AZURE_API_BASE")
    }
}, {
    "model_name": "gpt-3.5-turbo", 
    "litellm_params": { # params for litellm completion/embedding call 
[...]

Issue #47, opened 2023-07-05
Investigate Swap to GPT-4


FraserLee:
With falling API prices and rising importance of this project, I suspect this might now be possible. We should find the exact API costs of a message back and forth, work out some projected usage figures, and talk with Plex to see if it fits the org's budget.

ccstan99:
Worth investigating & keeping an eye on this but main issue before was latency. Again, if we have a good baseline then we can compare quality of generated output.

FraserLee:
I've used the new prompt-eng tooling I wrote to create a google sheet of answers comparing gpt-3.5-turbo to gpt-4. Imo the difference is so great that we have no choice but to switch, assuming the cost isn't prohibitively. I'm curious to know what others think
https://docs.google.com/spreadsheets/d/1vsaTHUVCBc0giH5UDD42R8eh2kMjvtQkIj9MKJMS5do/edit#gid=1376937604

FraserLee:
I've created a benchmark to compare outputs from gpt-3.5-turbo-16k to gpt-4.  We'll need a handful of people to go through and rank their preferences.
https://forms.gle/P1RnHh5ziVJkgAi99

ccstan99:
This is related to #34 so I'm going to close this as a duplicate.

PR #99, opened 2023-09-13
Add Budget Manager + Support for Anthropic, Cohere, Palm (100+ LLMs using LiteLLM)

ishaan-jaff:
Addressing: 
https://github.com/StampyAI/stampy-chat/issues/47 
https://github.com/StampyAI/stampy-chat/issues/55 

This PR addresses two problems:
- Add support for 100+ LLMs
- Use a budget manager for limiting $ spend per session or per user 

## Add support for 100+ LLMs
using LiteLLM https://github.com/BerriAI/litellm/ 
LiteLLM is a lightweight package to simplify LLM API calls - use any llm as a drop in replacement for gpt-3.5-turbo. 

Example
```python
from litellm import completion
## set ENV variables
os.environ["OPENAI_API_KEY"] = "openai key"
os.environ["COHERE_API_KEY"] = "cohere key"
messages = [{ "content": "Hello, how are you?","role": "user"}]
# openai call
response = completion(model="gpt-3.5-turbo", messages=messages)

# cohere call
response = completion(model="command-nightly", messages)

```

## Use a budget manager for limiting $ spend per session or per user 
LiteLLM exposes a budget manager for each session/user 
- We init a budget manager 
- Check the budget manager before making completion calls
- Update with costs after making completion calls

### DIFF
--- a/api/src/stampy_chat/chat.py
+++ b/api/src/stampy_chat/chat.py
@@ -2,6 +2,8 @@
 from dataclasses import asdict
 from typing import List, Dict, Callable
 import openai
+import litellm
+import uuid
 import re
 import tiktoken
 import time
@@ -25,6 +27,9 @@ HISTORY_FRACTION = 0.25 # the (approximate) fraction of num_tokens to use for hi
 CONTEXT_FRACTION = 0.5  # the (approximate) fraction of num_tokens to use for context text before truncating
 
 ENCODER = tiktoken.get_encoding("cl100k_base")
+
+# initialize a budget manager to control costs for gpt-4/other llms
+budget_manager = litellm.BudgetManager(project_name="stamp_chat")
 
 DEBUG_PRINT = True
 
@@ -142,7 +147,7 @@ def construct_prompt(query: str, mode: str, history: List[Dict[str, str]], conte
 import time
 import json
 
-def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str, str]], k: int = STANDARD_K, log: Callable = print):
+def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str, str]], k: int = STANDARD_K, log: Callable = print, session_id: str = ""):
     try:
         # 1. Find the most relevant blocks from the Alignment Research Dataset
         yield {"state": "loading", "phase": "semantic"}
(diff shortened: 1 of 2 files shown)


ishaan-jaff:
@FraserLee @henri123lemoine can i get a review on this pr ? 


if this initial commit looks good i can add docs/testing too

api/src/stampy_chat/chat.py
```diff
@@ -26,6 +28,9 @@
 
 ENCODER = tiktoken.get_encoding("cl100k_base")
 
+# initialize a budget manager to control costs for gpt-4/other llms
+budget_manager = litellm.BudgetManager(project_name="stampy_chat")
```
mruwnik, line 32: how does the budget per user get configured? Could you add a new item to `env.py` so that it can be configured? Also, what would the units be? (I had a very quick glance at the litellm docs, but otherwise don't know anything about it)

api/src/stampy_chat/chat.py
```diff
@@ -225,13 +235,16 @@ def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str,
 
 # convert talk_to_robot_internal from dict generator into json generator
 def talk_to_robot(index, query: str, mode: str, history: List[Dict[str, str]], k: int = STANDARD_K, log: Callable = print):
-    yield from (json.dumps(block) for block in talk_to_robot_internal(index, query, mode, history, k, log))
+    session_id = str(uuid.uuid4())
```
mruwnik, line 238: If I understand this, then budget manager has an internal dict to count how much a given session has used? But if you're creating a new id with each call to this function, then each session will have max 1 call? I'm planning on adding session ids, as it will be needed for logging anyway, so could you do this by extracting the session id from the request params in the `main.py` functions?

api/src/stampy_chat/chat.py
```diff
@@ -225,13 +235,16 @@ def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str,
...
+    budget_manager.create_budget(total_budget=10, user=session_id) # init $10 budget
```
mruwnik, line 239: don't hardcode it - create a setting in `env.py`

api/src/stampy_chat/chat.py
```diff
@@ -181,14 +186,19 @@ def talk_to_robot_internal(index, query: str, mode: str, history: List[Dict[str,
         t1 = time.time()
         response = ''
 
-        for chunk in openai.ChatCompletion.create(
+        # check if budget exceeded for session
+        if budget_manager.get_current_cost(user=session_id) <= budget_manager.get_total_budget(session_id):
```
mruwnik, line 190: is this for the number of allowed tokens or chat calls? Is it a hard total, or does it get reset every now and then? The code is run on gunicorn workers - how will that influence it, as I'm guessing litellm won't communicate across processes?

mruwnik (commented):
nice idea!

api/src/stampy_chat/chat.py
mruwnik: (I'll give it a proper look tomorrow)

PR #104, opened 2023-09-29
Tidy up

mruwnik:


api/src/stampy_chat/get_blocks.py
```diff
@@ -103,18 +100,18 @@ def get_top_k_blocks(index, user_query: str, k: int) -> List[Block]:
...
-        if not authors and match['metadata'].get('author'):
-            authors = [match['metadata'].get('author')]
+        authors = metadata.get('authors')
+        if not authors and metadata.get('author'):
+            authors = [metadata.get('author')]
 
         blocks.append(Block(
-            id = match['metadata']['hash_id'],
-            title = match['metadata']['title'],
+            id = metadata.get('hash_id'),
```
Aprillion, line 108: `hash_id` is not required so you don't want to raise Exception for that one, while `title, url, text` remain required?

Aprillion (approved):

api/src/stampy_chat/get_blocks.py
mruwnik: yup. It's not required for displaying, and should be provided in the newer ARD items. This is mainly a left over from the previous dataset, which could trip someone up. It *should* always be provided (famous last words...)

PR #106, opened 2023-09-29
Pytest

mruwnik:


.github/workflows/pytest.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@v3
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v3
+        with:
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          cd api
+          python -m pip install --upgrade pip
```
Aprillion, line 19: (minor) this feels unnecessary just to install `pipenv` TBH

api/src/stampy_chat/chat.py
```diff
@@ -29,117 +29,131 @@
...
+    if mode not in PROMPT_MODES:
+        raise ValueError("Invalid mode: " + mode)
 
+    # History takes the format: history=[
+    #     {"role": "user", "content": "Die monster. You don’t belong in this world!"},
+    #     {"role": "assistant", "content": "It was not by my hand I am once again given flesh. I was called here by humans who wished to pay me tribute."},
+    #     {"role": "user", "content": "Tribute!?! You steal men's souls and make them your slaves!"},
+    #     {"role": "assistant", "content": "Perhaps the same could be said of all religions..."},
+    #     {"role": "user", "content": "Your words are as empty as your soul! Mankind ill needs a savior such as you!"},
+    #     {"role": "assistant", "content": "What is a man? A miserable little pile of secrets. But enough talk... Have at you!"},
```
Aprillion, line 137: huh, where is this from? isn't it copyright infringement or something?

Aprillion (approved):

api/src/stampy_chat/chat.py
mruwnik: dunno. It was already here. But it looks like it's a (very) small part of a play or something, so should be fine as a case of free use

Issue #42, opened 2023-07-02
Same citations in different questions will have different numbers


tayler6000:
A paper may be cited in two questions but with different citation numbers.  Ideally, it should reuse the number but have a link to citation below both questions.

![image](https://github.com/StampyAI/stampy-chat/assets/931092/90181671-ea30-4341-ac78-28c4fd08341e)

I am marking this as an enhancement instead of a bug because this may be a current limitation with the API.

FraserLee:
This would be good, and doesn't need any API changes to pull off. Should be pretty easy, too. I ended up creating #46 after some initial investigation since there's some other work in this area we probably should be doing anyways, that would change how we go about implementing this.

PR #109, opened 2023-09-30
Same citation numbers

mruwnik:
https://github.com/StampyAI/stampy-chat/issues/42 Reuse citation numbers

image

### DIFF
--- a/api/src/stampy_chat/followups.py
+++ b/api/src/stampy_chat/followups.py
@@ -23,6 +23,9 @@ def search_authored(query: str):
 
 
 def get_followups(query):
+    if not query.strip():
+        return []
+
     url = 'https://nlp.stampy.ai/api/search?query=' + quote(query)
     response = requests.get(url).json()
     return [Followup(entry['title'], entry['pageid'], entry['score']) for entry in response]
--- a/web/src/components/citations.tsx
+++ b/web/src/components/citations.tsx
@@ -2,9 +2,7 @@ import type { Citation } from "../types";
 import { Colours, A } from "./html";
 
 
-// todo: memoize this if too slow.
-export const ProcessText: (text: string, base_count: number) => [string, Map<string, number>] = (text, base_count) => {
-
+export const formatCitations: (text: string) => string = (text) => {
   // ---------------------- normalize citation form ----------------------
   // the general plan here is just to add parsing cases until we can respond
   // well to almost everything the LLM emits. We won't ever reach five nines,
@@ -41,33 +39,29 @@ export const ProcessText: (text: string, base_count: number) => [string, Map<str
     /\[\s*([a-z]+)\s*\]/g,
     (_match: string, x: string) => `[${x}]`
   )
+  return response;
+}
 
-  // -------------- map citations from strings into numbers --------------
-
+export const findCitations: (text: string, citations: Citations[]) => Map<string, Citation> = (text, citations) => {
   // figure out what citations are in the response, and map them appropriately
-  const cite_map = new Map<string, number>();
-  let cite_count = 0;
+  const cite_map = new Map<string, Citation>();
 
   // scan a regex for [x] over the response. If x isn't in the map, add it.
   // (note: we're actually doing this twice - once on parsing, once on render.
   // if that looks like a problem, we could swap from strings to custom ropes).
   const regex = /\[([a-z]+)\]/g;
   let match;
-  let response_copy = ""
-  while ((match = regex.exec(response)) !== null) {
-    if (!cite_map.has(match[1]!)) {
-      cite_map.set(match[1]!, base_count + cite_count++);
+  while ((match = regex.exec(text)) !== null) {
+    const letter = match[1];
+    const citation = citations[letter.charCodeAt(0) - 'a'.charCodeAt(0)]
+    if (!cite_map.has(letter!)) {
+      cite_map.set(letter!, citation);
     }
-    // replace [x] with [i]
-    response_copy += response.slice(response_copy.length, match.index) + `[${cite_map.get(match[1]!)! + 1}]`;
   }
-
-  response = response_copy + response.slice(response_copy.length);
-
-  return [response, cite_map]
+  return cite_map
 }
 
-export const ShowCitation: React.FC<{citation: Citation, i: number}> = ({citation, i}) => {
+export const ShowCitation: React.FC<{citation: Citation}> = ({citation}) => {
 
   var c_str = citation.title;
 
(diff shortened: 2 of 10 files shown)


api/src/stampy_chat/followups.py
```diff
@@ -23,6 +23,9 @@ def search_authored(query: str):
 
 
 def get_followups(query):
+    if not query.strip():
+        return []
```
mruwnik, line 27: this was causing errors when the server returned an error (because of the empty query) which was parsed as followups

web/src/components/citations.tsx
```diff
@@ -79,25 +73,44 @@ export const ShowCitation: React.FC<{citation: Citation, i: number}> = ({citatio
...
-        ? citation.url
-        : `https://duckduckgo.com/?q=${encodeURIComponent(citation.title)}`;
-  return (
-    <A className={Colours[i % Colours.length] + " border-2 rounded text-sm no-underline w-min px-0.5 pb-0.5 ml-1 mr-0.5"}
-      href={url}>
-      [{i + 1}]
-    </A>
-  );
+export const CitationRef: React.FC<{citation: Citation}> = ({citation}) => {
+    const url = citation.url && citation.url !== ""
```
mruwnik, line 88: I keep wanting to add a linter, but worry that it'll mess up other PRs... :/

web/src/hooks/useSearch.ts
```diff
@@ -50,21 +51,22 @@ export async function* iterateData(res: Response) {
...
   var followups: Followup[] = [];
   for await (var data of iterateData(res)) {
     switch (data.state) {
       case "loading":
-        // display loading phases, once citations are available toss them
-        // into the current item.
+        setCurrent({ phase: data.phase, ...result });
+        break;
+
+      case "citations":
```
mruwnik, line 69: This adds a separate step for parsing citations

web/src/pages/index.tsx
```diff
@@ -70,6 +77,32 @@ const Home: NextPage = () => {
       }
   };
 
+
+  const updateCitations = (allCitations: Citation[], current: CurrentSearch) => {
```
mruwnik, line 81: The LLM flow is:
1. various loading stages
2. citations are passed in
3. more loading phases
4. the main response starts coming in in small chunks
5. more loading
6. the followups are fetched
7. done

The main issue here is that the citations are loaded first, and there usually are like 10 possible articles to be cited, but they're not used until they appear in the actual text. And most of them won't be used anyway.
This means that magic has to be done to work out which of the possible citations are actually cited, and for them to then get the appropriate numbers.

I dislike this function. Quite a bit. But couldn't come up with a cleaner way of doing it :/
At least most of the magic is focused here, with subsequent code assuming that each citation will have an `index` field with the canonical number to be displayed

web/src/hooks/useSearch.ts
```diff
@@ -74,11 +76,12 @@ export const extractAnswer = async (
...
       case "streaming":
         // incrementally build up the response
+        const content = formatCitations((result?.content || "") + data.content);
         result = {
+          content,
           role: "assistant",
-          content: (result?.content || "") + data.content,
           citations: result?.citations || [],
-          base_count: result?.base_count || baseReferencesIndex,
+          citationsMap: findCitations(content, result?.citations || []),
```
mruwnik, line 84: The LLM returns citations like "bla bla bla [a] ble ble [b][c]". These then need to be mapped to the appropriate numbers. This `citationMap` is a mapping of `{<letter>: <full citation object>}`. It's possible to change from letters to numbers - it's all in the LLM prompt

web/src/components/citations.tsx
```diff
@@ -79,25 +73,44 @@ export const ShowCitation: React.FC<{citation: Citation, i: number}> = ({citatio
   // if we don't have a url, link to a duckduckgo search for the title instead
   const url = citation.url && citation.url !== ""
         ? citation.url
-        : `https://duckduckgo.com/?q=${encodeURIComponent(citation.title)}`;
+        : `https://duckduckgo.com/?q=${encodeURIComponent(ndecitation.title)}`;
```
Aprillion, line 76: (major) is this typo?

web/src/components/citations.tsx
```diff
@@ -79,25 +73,44 @@ export const ShowCitation: React.FC<{citation: Citation, i: number}> = ({citatio
...
+    return (
+        <A className={Colours[(citation.index - 1) % Colours.length] + " border-2 rounded text-sm no-underline w-min px-0.5 pb-0.5 ml-1 mr-0.5"}
+           href={url}>
+            [{citation.index}]
+        </A>
+    );
 };
+
+
+export const CitationsBlock: React.FC<{text: string, citations: Map<string, Citation>, textRenderer: (t: str) => any}> = ({text, citations, textRenderer}) => {
```
Aprillion, line 100: (minor) is `React.FC` type used for consistency with the rest of the repo? just typing the props should be enough for function components, avoiding problems with implicit children - https://stackoverflow.com/a/59991281/1176601

web/src/components/citations.tsx
```diff
@@ -41,33 +39,29 @@ export const ProcessText: (text: string, base_count: number) => [string, Map<str
...
+export const findCitations: (text: string, citations: Citations[]) => Map<string, Citation> = (text, citations) => {
   // figure out what citations are in the response, and map them appropriately
-  const cite_map = new Map<string, number>();
-  let cite_count = 0;
+  const cite_map = new Map<string, Citation>();
 
   // scan a regex for [x] over the response. If x isn't in the map, add it.
   // (note: we're actually doing this twice - once on parsing, once on render.
   // if that looks like a problem, we could swap from strings to custom ropes).
   const regex = /\[([a-z]+)\]/g;
```
Aprillion, line 52: (minor) this supports only letters and skips numbers .. is that on purpose?

web/src/pages/index.tsx
```diff
@@ -43,9 +50,9 @@ function scroll30() {
 
 const Home: NextPage = () => {
   const [entries, setEntries] = useState<Entry[]>([]);
-  const [runningIndex, setRunningIndex] = useState(0);
   const [current, setCurrent] = useState<CurrentSearch>();
   const [sessionId, setSessionId] = useState()
+    const [citations, setCitations] = useState<Citation>([])
```
Aprillion, line 55: (minor) indentation looks funny on GH, is it aligned?

web/src/pages/index.tsx
```diff
@@ -70,6 +77,32 @@ const Home: NextPage = () => {
...
+        if (!citationsMapping[hash]) {
+          c.index = allCitations.length + 1;
+          allCitations.push(c);
+        } else {
+          c.index = citationsMapping[hash];
+        }
+      }
+    )
+    setCitations(allCitations)
+    setCurrent(current)
```
Aprillion, line 103: (minor) as a first impression, it looks suspicious to me to set both of these at the same event, I would expect that all citations need to be calculated less often than whatever is "current" ... but I didn't look into it in details, I'm happy if this works as intended

web/src/components/citations.tsx
mruwnik: good catch

web/src/components/citations.tsx
mruwnik: yes

web/src/components/citations.tsx
mruwnik: yes :/ There is a previous function that tries to clean up the text, but the idea is that the backend returns citations only as letters

web/src/pages/index.tsx
mruwnik: my editor assumes that it should be 4 spaces, not 2. This really should just be linted/autoformatted before pushing, but I haven't got round to it

web/src/pages/index.tsx
mruwnik: Totally agreed. This whole function is really ugly :/
This function should only be called when the current text gets updated. Which is quite a few times per query, but not that much.
`current` is the results of the current query, which gets updated a lot - once it's final it gets added to the history (which are previous interactions) and current is set to null (or something like that).
`allCitations` is the list of all displayed citations.
This function needs to update both or none - if the newest text chunk doesn't contain citations (or only known ones), then neither should be updated, but if it contains a new citation, then the map of `<letter>: <citation>` in the current object needs to be updated with the new indexes, but that also implies that the list of all show citations also needs to be updated with the new citations.

web/src/components/citations.tsx
Aprillion: in that case comments in other places in this repo about "this can be easily changed to numbers" should ideally point to this line :D

but I guess someone would notice if it stopped working after that change, so not a big deal

Aprillion (approved):

web/src/components/citations.tsx
mruwnik: "Easily" is "easily if you know exactly where to find the magical 3 lines and what to change" :D

PR #111, opened 2023-10-01
Chat options

mruwnik:
This one is a big boy. It adds something like OpenAI's playground for the chatbot:

image

I can split this into smaller PRs if this one is too large.
I'd also like to have it password protected or something, but thought that it's probably best to get the basic version out first

### DIFF
--- a/api/src/stampy_chat/chat.py
+++ b/api/src/stampy_chat/chat.py
@@ -6,92 +6,45 @@ from dataclasses import asdict
 from typing import List, Dict
 
 import openai
-import tiktoken
 
-from stampy_chat.env import COMPLETIONS_MODEL
+from stampy_chat import logging
 from stampy_chat.followups import multisearch_authored
 from stampy_chat.get_blocks import get_top_k_blocks, Block
-from stampy_chat import logging
+from stampy_chat.settings import Settings
 
 
 logger = logging.getLogger(__name__)
 
 
-STANDARD_K = 20 if COMPLETIONS_MODEL == 'gpt-4' else 10
-
-# parameters
-
-# NOTE: All this is approximate, there's bits I'm intentionally not counting. Leave a buffer beyond what you might expect.
-NUM_TOKENS = 8191 if COMPLETIONS_MODEL == 'gpt-4' else 4095
-TOKENS_BUFFER = 50  # the number of tokens to leave as a buffer when calculating remaining tokens
-HISTORY_FRACTION = 0.25 # the (approximate) fraction of num_tokens to use for history text before truncating
-CONTEXT_FRACTION = 0.5  # the (approximate) fraction of num_tokens to use for context text before truncating
-
-ENCODER = tiktoken.get_encoding("cl100k_base")
-
-SOURCE_PROMPT = (
-    "You are a helpful assistant knowledgeable about AI Alignment and Safety. "
-    "Please give a clear and coherent answer to the user's questions.(written after \"Q:\") "
-    "using the following sources. Each source is labeled with a letter. Feel free to "
-    "use the sources in any order, and try to use multiple sources in your answers.\n\n"
-)
-SOURCE_PROMPT_SUFFIX = (
-    "\n\n"
-    "Before the question (\"Q: \"), there will be a history of previous questions and answers. "
-    "These sources only apply to the last question. Any sources used in previous answers "
-    "are invalid."
-)
-
-QUESTION_PROMPT = (
-    "In your answer, please cite any claims you make back to each source "
-    "using the format: [a], [b], etc. If you use multiple sources to make a claim "
-    "cite all of them. For example: \"AGI is concerning [c, d, e].\"\n\n"
-)
-PROMPT_MODES = {
-    'default': "",
-    "concise": (
-        "Answer very concisely, getting to the crux of the matter in as "
-        "few words as possible. Limit your answer to 1-2 sentences.\n\n"
-    ),
-    "rookie": (
-        "This user is new to the field of AI Alignment and Safety - don't "
-        "assume they know any technical terms or jargon. Still give a complete answer "
-        "without patronizing the user, but take any extra time needed to "
-        "explain new concepts or to illustrate your answer with examples. "
-        "Put extra effort into explaining the intuition behind concepts "
-        "rather than just giving a formal definition.\n\n"
-    ),
-}
-
-# --------------------------------- prompt code --------------------------------
-
-
-
 # limit a string to a certain number of tokens
-def cap(text: str, max_tokens: int) -> str:
+def cap(text: str, max_tokens: int, encoder) -> str:
     if max_tokens <= 0:
         return "..."
 
-    encoded_text = ENCODER.encode(text)
+    encoded_text = encoder.encode(text)
 
     if len(encoded_text) <= max_tokens:
         return text
-    return ENCODER.decode(encoded_text[:max_tokens]) + " ..."
+    return encoder.decode(encoded_text[:max_tokens]) + " ..."
 
 
 Prompt = List[Dict[str, str]]
 
 
-def prompt_context(source_prompt: str, context: List[Block], max_tokens: int) -> str:
-    token_count = len(ENCODER.encode(source_prompt))
+def prompt_context(context: List[Block], settings: Settings) -> str:
+    source_prompt = settings.source_prompt_prefix
+    max_tokens = settings.context_tokens
+    encoder = settings.encoder
+
+    token_count = len(encoder.encode(source_prompt))
 
     # Context from top-k blocks
     for i, block in enumerate(context):
         block_str = f"[{chr(ord('a') + i)}] {block.title} - {','.join(block.authors)} - {block.date}\n{block.text}\n\n"
-        block_tc = len(ENCODER.encode(block_str))
+        block_tc = len(encoder.encode(block_str))
 
         if token_count + block_tc > max_tokens:
-            source_prompt += cap(block_str, max_tokens - token_count)
+            source_prompt += cap(block_str, max_tokens - token_count, encoder)
             break
         else:
             source_prompt += block_str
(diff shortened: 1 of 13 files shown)


api/src/stampy_chat/chat.py
```diff
@@ -6,128 +6,80 @@
...
 from stampy_chat.followups import multisearch_authored
 from stampy_chat.get_blocks import get_top_k_blocks, Block
-from stampy_chat import logging
+from stampy_chat.settings import Settings
 
 
 logger = logging.getLogger(__name__)
 
 
-STANDARD_K = 20 if COMPLETIONS_MODEL == 'gpt-4' else 10
```
mruwnik, line 20: all of these are moved over into the Settings class to be able to pass them around

api/src/stampy_chat/chat.py
```diff
@@ -6,128 +6,80 @@
...
-    ),
-}
-
-# --------------------------------- prompt code --------------------------------
-
-
-
 # limit a string to a certain number of tokens
-def cap(text: str, max_tokens: int) -> str:
+def cap(text: str, max_tokens: int, encoder) -> str:
```
mruwnik, line 20: the other changes in this file are basically to get it to use the settings object, rather than various constants

api/src/stampy_chat/chat.py
```diff
@@ -186,22 +139,24 @@ def talk_to_robot_internal(index, query: str, mode: str, history: Prompt, sessio
...
 
         # 3. Run both the standalone query and the full prompt through
         # moderation to see if it will be accepted by OpenAI's api
         check_openai_moderation(prompt, query)
 
         # 4. Count number of tokens left for completion (-50 for a buffer)
-        max_tokens_completion = remaining_tokens(prompt)
+        max_tokens_completion = remaining_tokens(prompt, settings)
+        if max_tokens_completion < 40:
+            raise ValueError(f"{max_tokens_completion} tokens left for the actual query after constructing the context - aborting, as that's not going to be enough")
```
mruwnik, line 151: should this be changed? I arbitrarily put 40, but 100 would probably also be too few

api/src/stampy_chat/get_blocks.py
```diff
@@ -144,13 +144,3 @@ def get_top_k_blocks(index, user_query: str, k: int) -> List[Block]:
     logger.debug(f'Time to get top-k blocks: {t2-t1:.2f}s')
 
     return join_blocks(blocks)
-
-
-# we add the title and authors inside the contents of the block, so that
-# searches for the title or author will be more likely to pull it up. This
-# strips it back out.
-def strip_block(text: str) -> str:
```
mruwnik, line 152: this continuosly raises warnings that it can't find the titles. Is it even used anymore? Wans't that the whole point of adding the other fields?

web/src/components/chat.tsx
```diff
@@ -0,0 +1,155 @@
+import { useState, useEffect } from "react";
```
mruwnik, line 1: I moved the actual chat functionality/component bits to a separate component so it can be imported easier. This also will make it easier to move it over to the stampy-ui repo...

web/src/components/searchbox.tsx
```diff
@@ -36,7 +36,8 @@ const SearchBoxInternal: React.FC<{search: (
     disable: () => void,
     enable: (f_set: Followup[] | ((fs: Followup[]) => Followup[])) => void,
   ) => void,
-}> = ({search}) => {
+  onQuery?: (q: string) => any,
```
mruwnik, line 39: adds a callback on changes to the search input. It's used to update the prompt preview

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+import Link from "next/link";
+
+import { queryLLM, getStampyContent, runSearch } from "../hooks/useSearch";
+import type { Mode, Entry, LLMSettings } from "../types";
+import Header from "../components/header";
+import Chat from '../components/chat';
+import { Controls } from "../components/controls";
+
+const MAX_FOLLOWUPS = 4;
+const DEFAULT_PROMPTS = {
```
mruwnik, line 14: I considered fetching these from the server, but couldn't be bothered to. It's a potential security risk, but not really that big of one, considering the code is available anyway... The downside of having it here is duplication and so the possibility of things getting out of sync. It's shouldn't be an issue, especially as this is pretty much just for testing

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+            "This user is new to the field of AI Alignment and Safety - don't " +
+            "assume they know any technical terms or jargon. Still give a complete answer " +
+            "without patronizing the user, but take any extra time needed to " +
+            "explain new concepts or to illustrate your answer with examples. " +
+            "Put extra effort into explaining the intuition behind concepts " +
+            "rather than just giving a formal definition.\n\n"
+        ),
+    }
+}
+const DEFAULT_SETTINGS = {
```
mruwnik, line 50: these are the current values on the backend

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+    completions: 'gpt-3.5-turbo',
+    encoder: 'cl100k_base',
+    topKBlocks: 10,         //  the number of blocks to use as citations
+    numTokens: 4095,
+    tokensBuffer: 50,      //  the number of tokens to leave as a buffer when calculating remaining tokens
+    maxHistory: 10,        //  the max number of previous items to use as history
+    historyFraction: 0.25, //  the (approximate) fraction of num_tokens to use for history text before truncating
+    contextFraction: 0.5,  //  the (approximate) fraction of num_tokens to use for context text before truncating
+}
+const COMPLETION_MODELS = ['gpt-3.5-turbo', 'gpt-4']
```
mruwnik, line 62: more can be added - these are the only ones I saw mentioned in the code

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+type InputFields = {
+    field: string,
+    label: string,
+    min?: number,
+    max?: number,
+    parser?: (v: string) => number,
+}
+
+const ChatSettings = ({settings, updateSettings}: ChatSettingsParams) => {
```
mruwnik, line 88: image

Basically a list of possible numerical values

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+}
+
+type ChatPromptParams = {
+    settings: LLMSettings,
+    query: string,
+    history: Entry[],
+    updateSettings: (updater: (settings: LLMSettings) => LLMSettings) => void,
+}
+
+const ChatPrompts = ({settings, query, history, updateSettings}: ChatPromptParams) => {
```
mruwnik, line 166: Displays the prompt that will be used. This is pretty much what will be sent to the LLM, apart from the citations chunks which first have to be fetched. So this panel allows one to pretty much pass in arbitrary prompts:

image

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,253 @@
...
+                <div>(This is where sources will be injected)</div>
+                {history.length > 0 && (
+                <TextareaAutosize
+                    className="border border-gray w-full px-1"
+                    value={settings.prompts.source.suffix}
+                    onChange={updatePrompt('source', 'suffix')}
+                />
+                )}
+            </details>
+            {history.length > 0 && (
```
mruwnik, line 189: This will display the text of the previous interactions, as the last `maxHistory` items will be used:

web/src/pages/playground.tsx
```diff
@@ -0,0 +1,257 @@
...
+            </details>
+            )}
+            <details open>
+                <summary>Question prompt</summary>
+                <TextareaAutosize
+                    className="border border-gray w-full px-1"
+                    value={settings.prompts.question}
+                    onChange={updatePrompt('question')}
+                />
+                <TextareaAutosize
```
mruwnik, line 206: this is where the prompt specifying the user's level goes

api/src/stampy_chat/chat.py
chriscanal: maybe it should be a class or instance variable? Also, What is the trade off here? More tokens cost more per response potentially, but has a better chance of giving the user what we want?

api/src/stampy_chat/get_blocks.py
chriscanal: maybe just change the logic to `if r: return r.group(1) if r else text` ? no need for warning. I'm good to delete this too.

api/src/stampy_chat/settings.py
```diff
@@ -0,0 +1,147 @@
...
+    def set_completions(self, completions, numTokens=None, topKBlocks=None):
+        self.completions = completions
+
+        # Set the max number of tokens sent in the prompt
+        if numTokens is not None:
+            self.numTokens = numTokens
+        elif completions == 'gtp-4':
+            self.numTokens = 8191
+        else:
+            self.numTokens = 4095
```
chriscanal, line 110: would be cool to drop the link to the api here in a comment: https://platform.openai.com/docs/models/gpt-4

web/src/types.ts
```diff
@@ -42,3 +42,21 @@ export type SearchResult = {
   result: Entry;
 };
 export type CurrentSearch = (AssistantEntry & { phase?: string }) | undefined;
+
+export type Mode = "rookie" | "concise" | "default";
+
+export type LLMSettings = {
+  prompts?: {
```
chriscanal, line 49: Are you sure all these settings should have the option of being undefined?

chriscanal (approved):
Epic pull request. Left a couple comments, nothing serious. Nice quality code here :)

api/src/stampy_chat/chat.py
mruwnik: There's already a `settings.numTokens` setting which should handle that. The problem here is that a whole lot of additional context is added to the prompt before it gets to this point. Specifially:
1. `settings.source_prompt_prefix`
2. up to `settings.topKBlocks` of citations from pinecone
3. `settings.source_prompt_suffix`
4. (optional) the last `settings.maxHistory` items from the conversation
5. `settings.question_prompt`
6.  (optional) `settings.mode_prompt` to regulate the complexity level of the answer
7. the actual query

This prompt can get quite big, taking up a lot of the `settings.numTokens` available tokens (in the worst case taking all of them), which limits the number left for the actual response. A seperate setting might help, but won't solve the underlying problem, which would probably require some creative book keeping to limit the number of tokens available for each step. Steps 1, 3, 5, 6 and 7 should always be sent to the LLM (but even then can take up all the tokens), so it's really only the context and history steps that should be limited.

web/src/types.ts
mruwnik: yes. They should have sane defaults in the API - they're mainly here to allow playing about with them, so no point in enforcing them all to be set

PR #112, opened 2023-10-02
fix typescript errors

mruwnik:
Adds linters and prettier. Only the commands for now + fixes, but that still is a lot...


Aprillion (approved):
commands look fine (copy&pasted from stampy-ui or something changed?)

FTR I only scrolled through fixes without checking too much 🤞

mruwnik:
from stampy-ui

PR #114, opened 2023-10-03
Chat options

mruwnik:
* Fix a couple of minor usability warts
* Add gpt-3.5-turbo-16k as a model choice
* Add an option to cancel queries
* Add prettier config (copied over from stampy ui)

### DIFF
--- a/web/src/components/chat.tsx
+++ b/web/src/components/chat.tsx
@@ -1,5 +1,5 @@
-import {useState, useEffect} from 'react'
-import {queryLLM, getStampyContent, runSearch} from '../hooks/useSearch'
+import { useState, useEffect } from "react";
+import { queryLLM, getStampyContent, runSearch } from "../hooks/useSearch";
 
 import type {
   CurrentSearch,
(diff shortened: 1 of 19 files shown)


mruwnik:
A lot of the changes are from prettier - if it's too annoying, I can separate the prettier stuff into a separate PR

web/src/components/chat.tsx
```diff
@@ -8,155 +8,147 @@ import type {
...
+        onNewEntry(entries)
       }
-      return entries;
-    });
-  };
+      return entries
+    })
+  }
 
   const search = async (
```
mruwnik, line 94: this function has a couple of changes - it accepts different arguments and has an AbortController + some changes in the handling of the search result

web/src/components/searchbox.tsx
```diff
@@ -1,135 +1,134 @@
...
-}> = ({ search, onQuery }) => {
-  const initial_query =
-    initialQuestions[Math.floor(Math.random() * initialQuestions.length)] || "";
+    query_source: 'search' | 'followups',
+    enable: (f_set: Followup[] | ((fs: Followup[]) => Followup[])) => void,
+    controller: AbortController
+  ) => void
+  onQuery?: (q: string) => any
+}> = ({search, onQuery}) => {
+  const initial_query = initialQuestions[Math.floor(Math.random() * initialQuestions.length)] || ''
```
mruwnik, line 40: this component has the largest functional changes

web/src/hooks/useSearch.ts
```diff
@@ -55,121 +61,132 @@ export const extractAnswer = async (
...
-  history: HistoryEntry[]
-): Promise<Response> =>
-  fetch(API_URL + "/chat", {
-    method: "POST",
-    cache: "no-cache",
+  history: HistoryEntry[],
+  controller: AbortController
+): Promise<Response | void> =>
+  fetch(API_URL + '/chat', {
+    signal: controller.signal,
```
mruwnik, line 117: this will cause the `fetch` to be aborted if `controller.abort()` is called. Which can be done via the `searchbox` button. A nice thing about this is that it also works while iterating over the response data, so all that was really needed in this file was to add this line here (and for the stampy request) and add some error handling

web/src/pages/semantic.tsx
```diff
@@ -1,95 +1,102 @@
...
+    enable: (f_set: Followup[]) => void,
+    controller: AbortController
   ) => {
-    disable();
-
-    const res = await fetch(API_URL + "/semantic", {
-      method: "POST",
+    const res = await fetch(API_URL + '/semantic', {
+      method: 'POST',
+      signal: controller.signal,
```
mruwnik, line 25: ditto here with the AbortController

mruwnik:
I removed the updated prettier stuff - too much bother for now

mruwnik:
ping

web/src/pages/playground.tsx
```diff
@@ -207,33 +239,50 @@ const ChatSettings = ({ settings, updateSettings }: ChatSettingsParams) => {
...
         min="0"
-        max={settings.tokensBuffer}
+        max={settings.numTokens}
+        updater={updateNum("tokensBuffer")}
       />
 
       <SectionHeader text="Prompt options" />
       <NumberInput
+        value={settings.topKBlocks}
         field="topKBlocks"
```
Aprillion, line 262: it there a point of `field` when passing value and updater manually?

Aprillion (approved):

web/src/pages/playground.tsx
mruwnik: it keeps the label and styleing in one place

PR #115, opened 2023-10-05
Move to langchain

mruwnik:
Migrates to langchain, using langchains methods, structures etc. Which are a massive pain to get through, as the code is frankly rubbish, and the docs often lacking...
I can also do a version with just the basic abstractions that they provide. That should allow a lot of the magic to be removed.
Also no tests for now - I'll add them once it's decided whether this is worth the bother


api/src/stampy_chat/callbacks.py
```diff
@@ -0,0 +1,161 @@
...
+from langchain.callbacks.base import BaseCallbackHandler
+from langchain.schema import BaseMessage
+
+from stampy_chat import logging
+
+
+logger = logging.getLogger(__name__)
+
+
+class StampyCallbackHandler(BaseCallbackHandler):
```
chriscanal, line 15: nice clean pattern :)

chriscanal:
> Migrates to langchain, using langchains methods, structures etc. Which are a massive pain to get through, as the code is frankly rubbish, and the docs often lacking... I can also do a version with just the basic abstractions that they provide. That should allow a lot of the magic to be removed. Also no tests for now - I'll add them once it's decided whether this is worth the bother

This is a huge PR. Is most of this refactoring necessary to move to langchain? It seems like you were refactoring a little to clean things up a little too. If we decide not to go with langchain can we keep some of this work? Also, any advice for testing this locally?

api/src/stampy_chat/logging.py
```diff
@@ -42,17 +42,14 @@ def is_debug(self):
         return self.isEnabledFor(DEBUG)
 
     def interaction(self, session_id: str, query: str, response: str, history, prompt, blocks):
-        prompt = [i for i in prompt if i.get('role') == 'system']
-        prompt = prompt[0].get('content') if prompt else None
```
chriscanal, line 46: I like the cleaner prompt interface for this exact reason, good job.

api/tests/stampy_chat/test_chat.py
chriscanal, line 1: Is this being deleted because it is no longer being used, or because you didn't want to update it to the new prompt data structure (string instead of dict)?

chriscanal:
Amazing work! I'm going to pull down and play with the code later today. Exciting stuff :)

api/src/stampy_chat/chat.py
```diff
@@ -1,221 +1,216 @@
-import time
```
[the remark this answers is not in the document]
mruwnik: it is, really. The old file pretty much does:

1. fetch context, i.e. the n most similar items from pinecone + push this to the frontend
2. get the last 10 items from the chat history
3. construct a prompt by joining strings and injecting the context + history (this is what is shown in the playground on the left)
4. check the OpenAI moderation endpoint to see if everything is correct
5. send the prompt to the LLM and stream the resulting text to the frontend
6. once the LLM has finished, send the result + the query to the NLP endpoint to get similar stampy answers + push this to the frontend
7. done

The current file does the same, but with langchain mechanisms. I initially did it all in a different file, but once it was all working, I removed the old chat.py file and renamed the langchain one. So it does the same as the old file did, pretty much in the same order, but in a totally new way

api/tests/stampy_chat/test_chat.py
mruwnik: both :D The old tests were for the old flow, and I wanted to get people's opinions on whether this even makes sense before spending a load of time writing tests for it

mruwnik:
Most of the changes were needed to get langchain to work. The refactor of `main.py` is sort of needed, but could have been left the same. But this pretty much totally replaces the previous code

api/src/stampy_chat/settings.py
```diff
@@ -130,26 +139,31 @@ def prompt_modes(self):
...
 
     @property
     def history_tokens(self):
         """The max number of tokens to be used for the history"""
-        return int(self.numTokens * self.historyFraction)
+        return int(self.numTokens * self.historyFraction) - len(self.encoder.encode(self.history_prompt))
+
+    @property
+    def max_response_tokens(self):
+        return self.numTokens - self.context_tokens - self.history_tokens
```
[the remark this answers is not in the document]
mruwnik: that's correct. And a good idea :D

mruwnik:
A version which doesn't use so much of the langchain magic is here: https://github.com/StampyAI/stampy-chat/pull/116

mruwnik:
Added tests - might as well do langchain properly. It's relatively easy to revert

PR #120, opened 2023-10-17
Allow chat items to be deleted

mruwnik:
I was uncertain whether to just delete any removed history items and totally forget about them, or to just soft delete them. The hard delete is nice in that it's a lot simpler, but has the downside that it would totally mess up the history which gets logged to the database. I considered having a proper changelog type thingy, which would allow one to recreate the whole session, but decided against it as it would get very complicated. On the other hand, the current logging implementation adds a number saying which interaction this is, based on the history length. An alternative to this would be to always query the database before writing, to get the number of interactions in this session.
I went with a soft delete, as that can be used later on to make an interactive session viewer, which is something I wanted to do anyway, to play about with seeing how different prompts and settings effect things.

image

### DIFF
--- a/api/src/stampy_chat/chat.py
+++ b/api/src/stampy_chat/chat.py
@@ -175,7 +175,7 @@ def make_memory(settings, history, callbacks):
         return_messages=True,
         callbacks=callbacks
     )
-    memory.set_messages(history)
+    memory.set_messages([i for i in history if i.get('role') != 'deleted'])
     return memory
 
 
@@ -192,16 +192,6 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
     if callback:
         callbacks += [BroadcastCallbackHandler(callback)]
     chat_model = get_model(streaming=True, callbacks=callbacks, max_tokens=settings.max_response_tokens)
-
-    memory = LimitedConversationSummaryBufferMemory(
-        llm=get_model(),
-        max_token_limit=settings.history_tokens,
-        max_history=settings.maxHistory,
-        chat_memory=ChatMessageHistory(),
-        return_messages=True,
-        callbacks=callbacks
-    )
-    memory.set_messages(history)
 
     chain = LLMChain(
         llm=chat_model,
--- a/api/src/stampy_chat/db/session.py
+++ b/api/src/stampy_chat/db/session.py
@@ -48,17 +48,15 @@ class ItemAdder:
         self._last_save = time.time()
 
     def commit(self):
-        with Session(self.engine) as session:
-            try:
+        try:
+            with Session(self.engine) as session:
                 session.add_all(self.batch)
                 session.commit()
                 logger.debug('added %s items', len(self.batch))
-                self.batch = []
-            except SQLAlchemyError as e:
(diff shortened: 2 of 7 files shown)


api/src/stampy_chat/chat.py
```diff
@@ -175,7 +175,7 @@ def make_memory(settings, history, callbacks):
         return_messages=True,
         callbacks=callbacks
     )
-    memory.set_messages(history)
+    memory.set_messages([i for i in history if i.get('role') != 'deleted'])
```
mruwnik, line 178: Make sure the LLM only sees non deleted items

api/src/stampy_chat/db/session.py
```diff
@@ -48,17 +48,15 @@ def __init__(self, engine=None, batch_size=100, save_every=1):
...
                 session.commit()
                 logger.debug('added %s items', len(self.batch))
-                self.batch = []
-            except SQLAlchemyError as e:
-                logger.warn('Got error when trying to commit to database: %s', e)
-                session.rollback()
-                raise e
+            self.batch = []
             self._last_save = time.time()
+        except SQLAlchemyError as e:
```
mruwnik, line 58: there's an interesting Heisenbug with sessions being dropped - this won't fix it, but at least won't break the chat when it happens. It only happens when the chatbot isn't used for a long period of time, which hopefully shouldn't be an issue once things start being used?
An objection can be raised that this will result in interactions not getting logged, which is an issue, but not a pressing one, as this is only for logging purposes, and should only happen if the error occurs and then the worker is shut down - otherwise the interactions will be saved the next time something happens, as they're saved in batches anyway

web/src/components/chat.tsx
```diff
@@ -164,9 +164,26 @@ const Chat = ({ sessionId, settings, onQuery, onNewEntry }: ChatParams) => {
...
+                className="delete-item absolute right-5 hidden cursor-pointer group-hover:block"
+                onClick={() => {
+                  const entry = entries[i];
+                  if (entry !== undefined) {
+                    entry.deleted = true;
+                    setEntries([...entries]);
+                  }
+                }}
+              >
+                X
```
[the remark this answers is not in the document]
Aprillion: ✕ == `✕` unicode character also exists if you want square X

web/src/hooks/useSearch.ts
```diff
@@ -223,7 +224,7 @@ export const runSearch = async (
     const history = entries
       .filter((entry) => entry.role !== "error")
       .map((entry) => ({
-        role: entry.role,
+        role: (entry.deleted ? "deleted" : entry.role) as EntryRole,
```
Aprillion, line 227: is there a reason to recycle `role` and not add a field `deleted`?

mruwnik: not really. I initially added an extra field, but this has the nice property that it will raise an error if not handled properly, as langchain won't accept an unknown role

web/src/components/chat.tsx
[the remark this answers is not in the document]
mruwnik: yup, that's nicer

Aprillion (approved):

PR #121, opened 2023-10-18
LLM settings from url

mruwnik:


api/main.py
```diff
@@ -44,20 +44,19 @@ def semantic():
...
-        return run_query(session_id, query, history, settings, callback)
+    settings = request.json.get('settings', {})
 
     def formatter(item):
         if isinstance(item, Exception):
             item = {'state': 'error', 'error': str(item)}
         return json.dumps(item)
 
+    def run(callback):
+        return run_query(session_id, query, history, Settings(**settings), callback)
```
mruwnik, line 58: the `Settings` class have various validators that can raise exceptions. This causes the request to fail with a uninformative (for the user) error, so make sure that a proper error instance is streamed back

api/src/stampy_chat/chat.py
```diff
@@ -191,7 +191,12 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
     callbacks = [LoggerCallbackHandler(session_id=session_id, query=query, history=history)]
     if callback:
         callbacks += [BroadcastCallbackHandler(callback)]
-    chat_model = get_model(streaming=True, callbacks=callbacks, max_tokens=settings.max_response_tokens)
+    chat_model = get_model(
+        streaming=True,
+        callbacks=callbacks,
+        max_tokens=settings.max_response_tokens,
+        model=settings.completions
```
mruwnik, line 198: the model setting wasn't being used

web/src/components/html.tsx
```diff
@@ -30,3 +44,78 @@ export const A: React.FC<{
     <a className={className}>{children}</a>
   );
 };
+
+const between =
```
mruwnik, line 48: these are just moved over from the `playground.tsx` file

web/src/components/settings.tsx
```diff
@@ -0,0 +1,309 @@
...
+export const makeSettings = (overrides: LLMSettings) =>
+  parseSettings(
+    Object.entries(overrides).reduce(
+      (acc, [key, val]) => updateIn(acc, key.split("."), val),
+      {}
+    ),
+    SETTINGS_PARSERS
+  );
+
+type ChatSettingsParams = {
```
mruwnik, line 133: the following lines are just moved over from `playground.tsx`

web/src/components/settings.tsx
```diff
@@ -0,0 +1,326 @@
...
+};
+export const MODELS = {
+  "gpt-3.5-turbo": { maxNumTokens: 4095, topKBlocks: 10 },
+  "gpt-3.5-turbo-16k": { maxNumTokens: 16385, topKBlocks: 30 },
+  "gpt-4": { maxNumTokens: 8192, topKBlocks: 20 },
+  /* 'gpt-4-32k': {maxNumTokens: 32768, topKBlocks: 30}, */
+};
+export const ENCODERS = ["cl100k_base"];
+
+/** Update the given `obj` so that it has `val` at the given path.
```
mruwnik, line 51: a bit of magic starts here - the idea is for each settings field to have a defined parser that will also return a default value if provided with an undefined value. Most of the magic is to handle the prompts, which are nested.
I considered just having separate keys for the prompt parts, but that would require a lot of messing around in other places, so I just kept the complexity here

web/src/pages/playground.tsx
```diff
@@ -1,387 +1,62 @@
...
+
   const setMode = (mode: [Mode, boolean]) => {
     if (mode[1]) {
       localStorage.setItem("chat_mode", mode[0]);
       updateSettings((settings) => ({ ...settings, mode: mode[0] }));
     }
   };
 
+  const changeSetting = (path: string[], value: any) => {
+    router.replace(
```
mruwnik, line 35: this will add any changes in the various widgets to the URL, so it'll be a lot easier to save/share parameters

mruwnik:
ping

Aprillion (approved):

PR #122, opened 2023-10-20
Test multiple questions

mruwnik:
Allows testing of all sample questions at once. Otherwise known as DDOS fun :D

image

### DIFF
--- a/web/src/pages/tester.tsx
+++ b/web/src/pages/tester.tsx
@@ -1,0 +1,169 @@
+import type { NextPage } from "next";
+import { useRouter } from "next/router";
+import { useState, useEffect, useCallback } from "react";
+import Page from "../components/page";
+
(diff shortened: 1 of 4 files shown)


web/src/pages/tester.tsx
```diff
@@ -0,0 +1,169 @@
...
+                  defaultChecked={selected}
+                  name={i.toString()}
+                  className="m-2"
+                />
+                <label htmlFor="i">{question}</label>
+              </div>
+            ))}
+            <button
+              onClick={() => {
+                controller.abort();
```
Aprillion, line 138: is this component fast to re-render between a double click? if not, the same controller aborted in the first click will be aborted again and the one created in first click won't be aborted at all .. you might need a `useRef` for a mutable reference that might change multiple times per render..

Aprillion (dismissed):

PR #123, opened 2023-10-31
Handle Discord messages properly

mruwnik:


api/src/stampy_chat/citations.py
```diff
@@ -63,7 +79,7 @@ def select_examples(self, input_variables: Dict[str, str]) -> List[dict]:
 def make_example_selector(k: int, **params) -> ReferencesSelector:
     embeddings = OpenAIEmbeddings(openai_api_key=OPENAI_API_KEY)
     vectorstore = Pinecone(PINECONE_INDEX, embeddings.embed_query, "hash_id", namespace=PINECONE_NAMESPACE)
-    return ReferencesSelector(vectorstore=vectorstore, **params)
+    return ReferencesSelector(vectorstore=vectorstore, k=k, **params)
```
Aprillion, line 82: why not remove `k` from parameter list on line 79 and only keep `**params` in both places?

Aprillion (approved):

api/src/stampy_chat/citations.py
mruwnik: good point :D Updated

Aprillion (approved):

PR #132, opened 2023-11-29
added langsmith code

jralduaveuthey:
added the code so that the interactions of the user with stampy in https://chat.stampy.ai/ are logged in langsmith to be able to monitor how the users interact with the chatbot

### DIFF
--- a/api/main.py
+++ b/api/main.py
@@ -10,7 +10,7 @@ from flask import Flask, jsonify, request, Response, stream_with_context
 from flask_cors import CORS, cross_origin
 
 from stampy_chat import logging
-from stampy_chat.env import PINECONE_INDEX, FLASK_PORT
+from stampy_chat.env import PINECONE_INDEX, FLASK_PORT, LANGCHAIN_API_KEY, LANGCHAIN_PROJECT
 from stampy_chat.settings import Settings
 from stampy_chat.chat import run_query
 from stampy_chat.callbacks import stream_callback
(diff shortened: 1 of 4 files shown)


api/main.py
```diff
@@ -94,6 +97,32 @@ def human(id):
     # <a href=\"https://stampy.ai/?state=6207&question=What%20is%20%22superintelligence%22%3F\">
     text = re.sub(r'<a href=\\"/\?state=(\d+.*)\\">', r'<a href=\"https://aisafety.info/?state=\1\\">', r.text)
 
+    LANGCHAIN_API_KEY = os.environ["LANGCHAIN_API_KEY"]
```
mruwnik, line 100: can you move these to `env.py` and import them from there?

api/src/stampy_chat/chat.py
```diff
@@ -18,6 +19,12 @@
 from stampy_chat.followups import StampyChain
 from stampy_chat.citations import make_example_selector
 
+from langsmith import Client
+if os.environ.get("LANGCHAIN_API_KEY"):
+    os.environ['LANGCHAIN_TRACING_V2'] = "true"
```
mruwnik, line 24: can these be explicitly passed into the `Client()` constructor, rather than playing about with env variable?

mruwnik (changes requested):

api/main.py
jralduaveuthey: modified it in new [commit](https://github.com/StampyAI/stampy-chat/pull/132/commits/d94abb7d9908728577beb7377a1749368d2860e1)

mruwnik (approved):

api/src/stampy_chat/chat.py
jralduaveuthey: modified it in new [commit](https://github.com/StampyAI/stampy-chat/pull/132/commits/d94abb7d9908728577beb7377a1749368d2860e1)

Now it raises an exception cause the setup of the client must be done with a valid key. So LANGCHAIN_TRACING_V2 sould not be "true" when there is no LANGCHAIN_API_KEY cause then the invokation of the chain does not work

Issue #136, opened 2024-01-14
Explore RAG adv methods & eval


ccstan99:
We might want to look into some new RAG chunk retrieval methods and evaluation metrics. This [short tutorial](https://www.deeplearning.ai/short-courses/building-evaluating-advanced-rag/) covers some interesting topics. They use LlamaIndex for implementation but we shouldn't need to. Looks like TruEra might have some good objective metrics to evaluate chunks and answers generated from the context. Some [notes & links to sample code](https://towering-elf-72b.notion.site/Building-and-Evaluating-Advanced-RAG-Applications-113ad9893177415fb8f789e7704303af).

Retrieval Methods:
- Sentence-window retrieval: chunk by sentence, returns sentence before & after semantic match
- Auto-merging retrieval: group set of relevant chunks into larger unit

Triad of metrics (to eval hyper-parameters & retrieval methods)
- Answer Relevance: Is the response relevant to the query?
- Context Relevance: Is the retrieved context relevant to the query?
- Groundedness: Is the response supported by the context?

wassname:
See also https://github.com/StampyAI/stampy-chat/issues/138

PR #141, opened 2024-05-20
added validation using tonic validate

varchanaiyer:
https://github.com/StampyAI/stampy-chat/assets/29728415/b7584d32-07b9-4d34-b1e1-68ab9d7ee2de

Demo video attached. For now, I am printing the output. But I can either push it to langsmith or to a db.

### DIFF
--- a/api/src/stampy_chat/chat.py
+++ b/api/src/stampy_chat/chat.py
@@ -1,5 +1,8 @@
 from typing import Any, Callable, Dict, List
 
+from tonic_validate import BenchmarkItem, ValidateScorer, LLMResponse
+from tonic_validate.metrics import AnswerConsistencyMetric, RetrievalPrecisionMetric, AugmentationAccuracyMetric
+
 from langchain.chains import LLMChain, OpenAIModerationChain
 from langchain_community.chat_models import ChatOpenAI
 from langchain_anthropic import ChatAnthropic
@@ -17,7 +20,7 @@ from stampy_chat.env import OPENAI_API_KEY, ANTHROPIC_API_KEY, LANGCHAIN_API_KEY
 from stampy_chat.settings import Settings, MODELS, OPENAI, ANTRHROPIC
 from stampy_chat.callbacks import StampyCallbackHandler, BroadcastCallbackHandler, LoggerCallbackHandler
 from stampy_chat.followups import StampyChain
-from stampy_chat.citations import make_example_selector
+from stampy_chat.citations import make_example_selector, get_top_k_blocks
 
 from langsmith import Client
 
@@ -322,6 +325,7 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
     :param Callable[[Any], None] callback: an optional callback that will be called at various key parts of the chain
     :returns: the result of the chain
     """
+
     callbacks = [LoggerCallbackHandler(session_id=session_id, query=query, history=history)]
     if callback:
         callbacks += [BroadcastCallbackHandler(callback)]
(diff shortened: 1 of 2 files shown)


api/src/stampy_chat/chat.py
```diff
@@ -340,9 +344,38 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
...
+        )
+
+    # Score the responses
+    scorer = ValidateScorer([
+        AnswerConsistencyMetric(),
+        AugmentationAccuracyMetric(),
+        RetrievalPrecisionMetric()
+    ])
+    run = scorer.score_responses([llm_response])
+    print(run.overall_scores)
```
mruwnik, line 375:
```
from stampy_chat import logging
logging.info(run.overall_scores)
```

api/src/stampy_chat/chat.py
```diff
@@ -340,9 +344,38 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
...
         memory=make_memory(settings, history, callbacks)
     )
+
     if followups:
         chain = chain | StampyChain(callbacks=callbacks)
+
     result = chain.invoke({"query": query, 'history': history}, {'callbacks': []})
+
+    #Validate results
+    contexts=[c['text'] for c in get_top_k_blocks(query, 5)]
```
mruwnik, line 354: you're querying the vectorstore twice for each request, which will make things a lot slower - how hard would it be to reuse the previously fetched examples?

api/src/stampy_chat/chat.py
```diff
@@ -340,9 +344,38 @@ def run_query(session_id: str, query: str, history: List[Dict], settings: Settin
...
+            benchmark_item=benchmark
+        )
+
+    # Score the responses
+    scorer = ValidateScorer([
+        AnswerConsistencyMetric(),
+        AugmentationAccuracyMetric(),
+        RetrievalPrecisionMetric()
+    ])
+    run = scorer.score_responses([llm_response])
```
mruwnik, line 374: is this actually returned to the user? If not, then I'd suggest doing it after notifying the frontend that everything is done, i.e. moving the `if callback (...)` clause before these

api/src/stampy_chat/chat.py
mruwnik: you can also send messages to discord with the logger

api/src/stampy_chat/chat.py
mruwnik, line 354: I'm guessing this shouldn't be executed for each query. How about adding a flag to the settings object?

mruwnik:
you have something wrong with the pipenv dependancies, which you'll have to fix. Did you add dependancies with `pipenv install <whatever>`?

api/src/stampy_chat/chat.py
varchanaiyer: Yes you are right. I had a lot of challenges in this part and it took the majority of my time with this ticket.

Unfortunately, I could not figure out how to get a hook into the LangChain Semantic Similarity Selector. Without a hook, there is no way to get the previously fetched examples. I also looked at the source code of RAGAs (another RAG validation framework) and saw that they too could not figure out how to put a hook into LangChain and stopped supporting it last year.

I tested the latency and saw that querying the vectorstore does not add much latency and unlike querying an LLM, there is no additional cost to it. This is why I decided to query the vector store again. I think we can put this in the backlog and monitor if LangChain updates their API

api/src/stampy_chat/chat.py
varchanaiyer: Makes sense. I will make this change. The reason I kept it this way is because I was worried that a "fast" user might ask the next question before the evaluation is done running. This might overwhelm the system and spawn a lot of processes.

I don't think this will be useful for our users. We may want to track it internally first and then decide what would be the best way to display this to users. For instance, I don't think users will know what "context-precision" is or care about it.

varchanaiyer:
> you have something wrong with the pipenv dependancies, which you'll have to fix. Did you add dependancies with `pipenv install <whatever>`?

Yes that is how I added dependencies. Is that the wrong way of doing it? I am not familiar with pipenv (have been using vanilla virtualenv all this time), so maybe I missed a step?

api/src/stampy_chat/chat.py
varchanaiyer: I think that this is required for each query because `tonic_validate`, checks to see if the LLM's answer is consistent with the context that was retrieved.

api/src/stampy_chat/chat.py
varchanaiyer: That would be really cool! I will add that in!

api/src/stampy_chat/chat.py
mruwnik: yes, but should it always do that, is the question. This makes the query slower, so for now I'd just use it for testing, rather than always running it

varchanaiyer: I get what you mean. I will add a flag to the settings object to prevent it from running all the time.

varchanaiyer:
https://github.com/StampyAI/stampy-chat/issues/136

PR #157, opened 2025-05-22
Update versions + add option to show full prompt

mruwnik:


web/src/styles/globals.css
```diff
@@ -120,3 +120,14 @@ ol {
 .text-section {
   padding-bottom: 8px;
 }
+
```
LeMurphant, line 123: I'm confused, isn't that managed in stampy-ui? c.f. https://github.com/StampyAI/stampy-ui/pull/918

mruwnik: that too, but this is for [the playground](https://chat.stampy.ai/playground)

LeMurphant (approved):
Looks good to me, although I could not test it since my stampy-chat local copy does not work.

PR #161, opened 2025-07-19
Replace prompts with cautious ones based on MIRI documents

nerualauren:


api/src/stampy_chat/citations.py
```diff
@@ -93,3 +86,13 @@ def retrieve_docs(
 
 def get_top_k_blocks(query: str, k: int) -> list[Block]:
     return retrieve_docs(query, [], Settings())[:k]
+#+def fix_text(received_text):
```
mruwnik, line 89: this should probably be removed?

api/src/stampy_chat/llms.py
```diff
@@ -44,33 +65,90 @@ def call_anthropic(
...
+    else:      return response.choices[0].message.content
+
+def openai_stream(response):
     for event in response:
         if event.type == "response.output_text.delta":
             yield LLMChunk(type="response", text=event.delta)
 
 
-def query_llm(prompt: str, settings: Settings) -> Generator[LLMChunk, None, None]:
+def call_google(
```
mruwnik, line 89: nice

api/src/stampy_chat/prompts.py
```diff
@@ -18,34 +28,112 @@ def truncate_history(history: list[Message], max_tokens: int) -> list[Message]:
...
+
+    mode = ""
+
+    last_parts = []
+    if settings.completions.startswith('google'):
+        modelname = 'Gemini'
+    elif settings.completions.startswith('anthropic'):
+        modelname = 'Claude'
+    elif settings.completions.startswith('openai'):
+        modelname = 'GPT'
```
mruwnik, line 120: [nit] I'd prefer this be a property in the settings object

api/src/stampy_chat/settings.py
```diff
@@ -108,17 +124,21 @@ def __init__(
...
         maxHistory=10,
         maxHistorySummaryTokens=200,
+        hyde_max_tokens=100,
         historyFraction=0.25,
         contextFraction=0.5,
         **_kwargs,
     ) -> None:
         self.prompts = prompts
         self.mode = mode
+        #import pudb; pudb.set_trace()
```
mruwnik, line 140: remove?

api/stampy_client.py
```diff
@@ -0,0 +1,859 @@
+#!/usr/bin/env python3
```
mruwnik, line 1: I don't like having scripts like this commited - they tend to rot. Isn't the frontend sufficient for general testing? i.e. I'm all for you having your own scripts - this is the way - but dunno about commiting them

api/test_client.py
```diff
@@ -0,0 +1,65 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
mruwnik: yup, claude likes to do that

mprocs.yaml
```diff
@@ -0,0 +1,21 @@
+procs:
```
mruwnik, line 1: what's this for?

prompts/detailed-cautious-epistem-safetyinfo-v1-2507140045-aa7693.txt
```diff
@@ -0,0 +1,108 @@
+`
```
mruwnik, line 1: dunno if these prompts should be part of the codebase. They seem like they should live somewhere else?

PR #169, opened 2025-08-07
Fix citations, thinking. Pls merge as separate commits

nerualauren:


web/package.json
```diff
@@ -32,6 +32,7 @@
     "autoprefixer": "^10.4.7",
     "eslint": "^8.34.0",
     "eslint-config-next": "^13.2.1",
+    "patch-package": "^8.0.0",
```
mruwnik, line 35: what's this for?

web/src/components/citations.tsx
```diff
@@ -149,7 +149,7 @@ export const CitationsBlock: React.FC<{
   citations: Map<string, Citation>;
   textRenderer: (t: string) => any;
 }> = ({ text, citations, textRenderer }) => {
-  const regex = /\[([a-z]+)\]/g;
+  const regex = /\[([a-z0-9]+)\]/g;
```
mruwnik, line 152: they should be either numbers or letters. Doesn't seem a good idea to mix them

nerualauren: for now though this makes the citation links work again in the playground

PR #170, opened 2025-08-07
Handle bad messages

mruwnik:


api/main.py
```diff
@@ -45,6 +46,18 @@ def semantic():
 # ------------------------------------ chat ------------------------------------
 
 
+def clean_history(history: list[Message]) -> list[Message]:
```
mruwnik, line 49: might as well attempt to fix things if they're sent incorrectly

mruwnik: it's not any more - there was a frontend bug (or 5) which was causing this. But it seems like a good default to have anyway

PR #188, opened 2025-09-19
new prompt

nerualauren:


api/src/stampy_chat/citations.py
```diff
@@ -102,7 +102,47 @@ def retrieve_docs(query: str, settings: Settings) -> list[Block]:
...
+            # New document
+            doc_key = (title, base_url)  # Use tuple as unique key
+            if title:
+                seen_titles[title] = doc_key
+            if base_url:
+                seen_urls[base_url] = doc_key
+            seen_docs[doc_key] = (score, match, reference_counter)
+            reference_counter += 1
+        elif score > seen_docs[doc_key][0]:
+            # Better score for existing document
```
LeMurphant, line 140: Maybe clarify "better score *than* existing document", I misread "existing is better".

LeMurphant (dismissed):

PR #209, opened 2026-09-21
Fix lost interaction logs and get_doc hash_id crash

nerualauren:
Prod backend journal, 30 days to 2026-09-21, showed three bugs; all reproduced in source and fixed here. Evidence and a wider status check (traffic, corpus freshness, two live answer transcripts) in `docs/status/2026-09-21-stampy-status.md`.

1. `interactions.query` was VARCHAR(1028): 201 turns failed to log ("Data too long"), so the DB shows zero interactions since Aug 24 while requests kept running. Widened `query`/`chunks` to LONGTEXT. **The ALTER is already applied on prod** (verified in information_schema); the migration is for the record and for local DBs.
2. Remote MySQL with no `pool_pre_ping`: first write after idle failed ("MySQL Connection not available", 24 rows lost). Added pre_ping + recycle.
3. `get_doc`'s docstring says "Accepts hash_id" but the parameter is `id`; Claude sent `hash_id=` and the uncaught TypeError killed the whole request (19 times). Added the alias; the tool wrapper now returns an `Error: ... Valid arguments: ...` string on bad kwargs so the model can retry.

Tests: `api/tests/stampy_chat/test_tools.py`, both fail on the old code; full suite 18 passed (`.venv/bin/python -m pytest tests`).

Merging triggers the production deploy (restarts backend/website/mcp).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_017TDp82vMNqDUzQiqdAt8t2


nerualauren:
Three sonnet reviewers, dispatched by perspective before merge. What they found and what changed:

**Correctness reader**
- Real regression: the new `TypeError` guard in the tool wrapper returned early without appending a tool event, and `chat.py` pairs events to `tool_use` ids by position, so a bad call next to a good one would mislabel the good one's result. Fixed: the error path now emits its event like any result; test asserts it.
- Real bug: the brief escaped `{}` as `{{}}`, but it is appended after both `format` passes, so titles with braces would reach the model doubled (and my test locked it in). Fixed; test now asserts literal braces.
- Clean: the streaming citation rewriter, end to end.

**Abuse reader**
- Cost was charged by query length while spend is driven by thinking and tool rounds. Now the answer's real cost (one unit per model call) is settled against both the per-IP and the global budget after the stream ends; global is a unit budget, not a request count.
- Dict keys never reclaimed (`repeats` keyed by full query text). Now pruned when touched and swept wholesale above 10k keys.
- Streaming refusals are HTTP 200, invisible to status monitoring. Added an `X-RateLimited: <reason>` header; the warning log line is the other signal.
- Accepted as known limits: one NAT shares one budget (all limits are `RL_*` env vars); the repeat check is a UX nudge, trivially mutated; `history` length is uncounted (truncated downstream).

**The answering model's reader** (read the rendered brief as the Sonnet who gets it)
- The rendered list is a wall of "model X hacked Y" titles arriving right after a prompt that names the reader among the minds at stake. Added a line: these are reports about specific named systems and incidents, not claims about you, read them like news of another profession's mistakes.
- List is date-sorted while the header said karma-ranked; header now says both. Added a line on forum attention skewing alarming. Cut a duplicate citability sentence. `ip_daily` message now mentions shared connections.

37 tests pass; five reintroduced bugs each fail exactly one test.

nerualauren:
Deployed 2026-09-21 07:42 UTC. The deploy-to-production action failed at the ssh handshake ("connection reset by peer", run 35574003144); the same reset hits interactive ssh on rapid reconnects, so likely a connection-rate limit on port 22 rather than the change. Deployed by hand: `/var/www/update.sh`, then website/backend/mcp restarts. Prod is on 36cdfc0.

Verified on prod after restart: GET /chat "has a warning shot happened yet" now reports the OpenAI/HuggingFace incident with sources (42 s); streaming POST from Origin aisafety.info renders `[TB](https://intelligence.org/briefing/)`; the interaction logged as row 3500; backend journal clean.

Pre-existing, not fixed: GET /chat/<q> passes session_id None and the logger raises "badly formed hexadecimal UUID string", so that test path never logs.

